PR #61, opened 2022-04-03
solution

vitaliy-zhyvyliak:
[DEMO](https://vitaliy-zhyvyliak.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,96 @@ const goodsFromServer: string[] = [
...
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState(goodsFromServer);
+  const [isVisible, setVisibility] = useState(false);
```
jursylian, line 19: можна поміняти `setVisibility` на `setIsVisible` , але як на мене і так непогано:)

src/App.tsx
```diff
@@ -14,11 +14,96 @@ const goodsFromServer: string[] = [
...
 
-export default App;
+  const showGoods = () => {
+    setVisibility(!isVisible);
+  };
```
Ihor-Karpyn, line 23: function has name showGoods, to show tou shoul set true. If it's toggle give function right name and use functional set state 
```suggestion
  const showGoods = () => {
    setVisibility(true);
  };
```

Ihor-Karpyn (approved):

PR #91, opened 2022-05-11
Develop

MariaBondar:
https://mariabondar.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,7 +18,7 @@ const App: React.FC = () => {
   const [isVisible, setIsVisible] = useState(true);
   const [goods, setGoods] = useState(goodsFromServer);
 
-  const notVisible = () => {
+  const toggleVisibiity = () => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,11 +14,73 @@ const goodsFromServer: string[] = [
...
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+const App: React.FC = () => {
+  const [isVisible, setIsVisible] = useState(true);
+  const [goods, setGoods] = useState(goodsFromServer);
+
+  const notVisible = () => {
```
VETALEK, line 21:
```suggestion
  const toggleVisibiity = () => {
```
confusing func name

VETALEK (changes requested):
Works just fine) Need only styles and a single rename

Andr1yK (approved):

PR #109, opened 2022-05-13
Develop

Yurii837:


Helexi (approved):
Good job, but recommend adding styles

src/App.tsx
```diff
@@ -14,11 +15,91 @@ const goodsFromServer: string[] = [
...
+
+  const prepGoods: string[] = sortBy !== ''
+    ? [...goods].sort((good1, good2) => {
+      switch (sortBy) {
+        case 'alphab': return good1.localeCompare(good2);
+        case 'length': return good1.length - good2.length;
+        default: return 0;
+      }
+    })
+    : [...goods];
```
MaksimusCRS, line 49: Just add some logic to default case

MaksimusCRS (changes requested):
Add some styles before review

src/App.tsx
Yurii837: sudenly 0 is logic(array did not sort)

MaksimusCRS: You don't need ternary operator here, just pass correct callback in default case

MaksimusCRS (approved):

PR #111, opened 2022-05-13
react list of goods hooks

RuslanaTomnyuk:
[DEMO LINK](https://RuslanaTomnyuk.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -14,11 +14,86 @@ const goodsFromServer: string[] = [
...
+          </button>
+        </div>
+      </div>
+    )
+    : (
+      <button
+        type="button"
+        className="button"
+        onClick={() => setStartVisible(true)}
+      >
```
HelenPetrovska, line 93: you can write function, but it's your option)

HelenPetrovska (approved):
Great!

src/App.tsx
AlinaMatuschak: Agree with @HelenPetrovska , better create separate method

AlinaMatuschak (approved):

PR #159, opened 2022-06-07
Task Done

khozhainov-aleksandr:
[DEMO LINK](https://khozhainov-aleksandr.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import './App.css';
+import React, { useState } from 'react';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
 
-export default App;
+  const isVisibleList = () => {
+    setVisibleList(!visibleList);
+  };
+
+  const makeReverse = () => {
+    setGoods([...goods].reverse());
+  };
+
```
igoree88, line 28:
```suggestion
const makeReverse = () => {
    setGoods([...goods].reverse());
  };
    const makeSortAlphabet = () => {
    setGoods([...goods].sort((a, b) => a.localeCompare(b)));
  };

  const makeSortLength = () => {
    setGoods([...goods].sort((a, b) => a.length - b.length));
  };
```
If you make some changes depending on the value of previous state, you should use functional setState:
 setGoods((currentGoods) => ({...}))

src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
+  const makeReset = () => {
+    setGoods(goodsFromServer);
+  };
+
+  return (
+    <div className="App">
+      <h1 className="title">Goods</h1>
+
+      {!visibleList && (
+        <button
```
igoree88, line 46:
```suggestion
      {!visibleList ? ( ) : ( )}
        
```
In this case maybe it will better to use  ternary operator. I think  it can little bit improve the readability of the code.

igoree88 (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #160, opened 2022-06-08
Solution React list of goods hooks

dmytrosmnnk:
https://dmytrosmnnk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/components/GoodsList/GoodsList.tsx
+++ b/src/components/GoodsList/GoodsList.tsx
@@ -1,0 +1,95 @@
+import React, { useState } from 'react';
+import './GoodsList.scss';
+
+type Props = {
+  goods: string[];
+};
+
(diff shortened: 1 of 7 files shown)


src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,95 @@
...
+        return (g1.name.length - g2.name.length);
+
+      default:
+        return 0;
+    }
+  });
+
+  if (reverse) {
+    preparedGoods.reverse();
+  }
```
vkryvytskyy, line 35: For now reverse works only single time(then you need to apply sorting using another button), make it reversing the array each time user clicks on it

vkryvytskyy (changes requested):

PR #161, opened 2022-06-08
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_list-of-goods-hooks/)


khozhainov-aleksandr (approved):
Good job and good styles

src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+  const [visibleGoods, setVisibleGoods] = useState([...goodsFromServer]);
+
+  const findIndexOfName = (currentName: string): number => (
+    goodsFromServer.findIndex(name => name === currentName)
+  );
+
+  return (
+    <div className="conteiner">
+      <div className="app">
+        {!listVisible && (
```
Anna-Tkachenko, line 28: You can also use the next syntax
{listVisible
  ? ()
  : ()
}

src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+                </li>
+              ))}
+            </ul>
+
+            <button
+              className="button is-primary"
+              type="button"
+              onClick={() => (setVisibleGoods(
+                [...visibleGoods].reverse(),
+              ))}
```
Anna-Tkachenko, line 60: If you make some changes depending on the value of previous state, you should use function inside setVisibleGoods method (as we used functional setState)
setVisibleGoods((currentGoods) => ({...}))

src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+            >
+              Reverse
+            </button>
+            <button
+              className="button is-primary"
+              type="button"
+              onClick={() => (setVisibleGoods(
+                [...visibleGoods].sort((a, b) => a.localeCompare(b)),
```
Anna-Tkachenko, line 68: Same here

src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+              ))}
+            >
+              Sort alphabetically
+            </button>
+            <button
+              className="button is-primary"
+              type="button"
+              onClick={() => (setVisibleGoods(
+                [...visibleGoods].sort((a, b) => a.length - b.length),
```
Anna-Tkachenko, line 77: And here

Anna-Tkachenko (changes requested):

src/App.tsx
igoree88: Fixed

src/App.tsx
igoree88: You're absolutely right! Fixed.

src/App.tsx
igoree88: I agree, in this case maybe it will really improve the readability of the code. Fixed

PR #166, opened 2022-06-11
Develop

AndriySimyachenko:
[DEMO LINK](https://AndriySimyachenko.github.io/react_list-of-goods-hooks/)


alexandra-protyanova (approved):

src/components/List.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </ul>
+
+      <div>
+
+        <button
+          type="button"
+          onClick={() => {
+            reverseGoods();
+          }}
+        >
```
Anna-Tkachenko, line 58: Inner function is redundant here
```suggestion
          onClick={reverseGoods}
        >
```

src/components/List.tsx
```diff
@@ -0,0 +1,93 @@
...
+          onClick={reset}
+        >
+          Reset
+        </button>
+
+        <button
+          type="button"
+          onClick={() => {
+            setSortBy('length');
+            universalSort();
```
Anna-Tkachenko, line 83: Will be better just pass needed argument in universalSort function instead of storing in state object

Anna-Tkachenko (approved):

PR #171, opened 2022-06-13
Initial commit

amitpatiljc:
[DEMO LINK](https://amitpatiljc.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,11 +28,11 @@ const App: React.FC = () => {
   const [sortBy, setSortBy] = useState(SortBy.null);
 
   const start = () => {
-    setStart(current => !current);
+    setStart(prev => !prev);
   };
 
   const reverse = () => {
-    setReverse(current => !current);
+    setReverse(prev => !prev);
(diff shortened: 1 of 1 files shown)


andreisucman (approved):
Looks good.

src/App.tsx
```diff
@@ -14,11 +15,112 @@ const goodsFromServer: string[] = [
...
+}
+
+const App: React.FC = () => {
+  let goodsList = [...goodsFromServer];
+  const [isListVisible, setStart] = useState(false);
+  const [isReverse, setReverse] = useState(false);
+  const [sortBy, setSortBy] = useState(SortBy.null);
+
+  const start = () => {
+    setStart(prev => !prev);
```
meforoma, line 31: in case of `reverse` inverting (`!`) makes total sense,
but when start used only once and then the button is not visible,
so inverting does not fit, regular `true` would be just fine
yet it does not impact much

meforoma (approved):

PR #183, opened 2022-06-19
Solution

naziamoff:
[DEMO LINK](https://naziamoff.github.io/react_list-of-goods-hooks/)


vkryvytskyy (approved):
Would be nice to complete advanced tasks also

PR #192, opened 2022-07-11
solution

kostya-konev:
[DEMO LINK](https://kostya-konev.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+                  </option>
+                  <option value="8">
+                    8
+                  </option>
+                  <option value="9">
+                    9
+                  </option>
+                  <option value="10">
+                    10
+                  </option>
```
Bohdan-Maksymiuk, line 135: you could map this

src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+      {
+        !isShown
+          ? (
+            <button
+              type="button"
+              className="button"
+              onClick={() => setIsShown(true)}
+            >
+              Start
+            </button>
```
Bohdan-Maksymiuk, line 52: better do not use such small check in return and do not complicate structure of `return`
```suggestion
  const sortByLength = () => {
    setGoods([...goods].sort((a, b) => b.length - a.length));
  };
  
  if (!isShown) {
    return (your button here)
  }

  return (
    <div className="App">
      {
        !isShown
          ? (
            <button
              type="button"
              className="button"
              onClick={() => setIsShown(true)}
            >
              Start
            </button>
```
You can do like this and remove conditional rendering here

src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+                {goods.filter(good => good.length >= +length)
+                  .map((product) => {
+                    return (
+                      <li key={product} className="App__item">
+                        <span className="App__span">
+                          {product}
+                        </span>
+                      </li>
+                    );
+                  })}
```
Bohdan-Maksymiuk, line 65:
```suggestion
                  .map((product) => (
                      <li key={product} className="App__item">
                        <span className="App__span">
                          {product}
                        </span>
                      </li>
                  ))}
```
use shorthand

Bohdan-Maksymiuk (changes requested):

SanyaBratashchuk (changes requested):
Add all files to the pr

ArturErmolenko (approved):

PR #216, opened 2022-07-17
initial commit

Kate-Khon:
[DEMO LINK](https://kate-khon.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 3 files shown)


SpaceHook (approved):
GJ👍

src/App.tsx
```diff
@@ -15,33 +14,131 @@ const goodsFromServer: string[] = [
...
+    element.length >= minLength
+  ));
+
+  if (sortType === SortType.ALPABET) {
+    visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+  } else if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((good1, good2) => good1.length - good2.length);
+  }
+
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
```
jstmpelowycz, line 39: perhaps, switch case could better fit it? leaving else-if-statement as the last one is not a common practice

src/App.tsx
```diff
@@ -15,33 +14,131 @@ const goodsFromServer: string[] = [
...
+              isReversed,
+              lengthValue,
+            ).map(product => (
+              <li
+                className="Goods__item"
+                key={product}
+              >
+                {product}
+              </li>
+            ))}
```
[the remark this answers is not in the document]
jstmpelowycz: I'd rather put it into a var, and the just map that var

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -15,33 +14,135 @@ const goodsFromServer: string[] = [
...
+  LENGTH = 'length',
+}
+
+function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+  minLength: number,
+) {
+  const visibleGoods = [...goods].filter(element => (
```
jstmpelowycz, line 29:
```suggestion
  const visibleGoods = goods.filter(element => (
```

jstmpelowycz (changes requested):

illiarerun (approved):

PR #222, opened 2022-07-17
Solution

AntonYegorov:
[DEMO LINK](https://AntonYegorov.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,6 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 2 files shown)


illavodonis (approved):

src/App.tsx
```diff
@@ -15,33 +14,112 @@ const goodsFromServer: string[] = [
...
+  };
+
+  const sortByLength = () => {
+    setIsSorted(true);
+    setSortBy('length');
+  };
+
+  const sortByAlphabetically = () => {
+    setIsSorted(true);
+    setSortBy('alphabet');
```
jstmpelowycz, line 40: use enum instead

jstmpelowycz:
```
setSortBy(SortType.ByAlphabet);
```
or something like that

jstmpelowycz (approved):

PR #229, opened 2022-07-18
solution

Victance:
[DEMO LINK](https://victance.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,33 +15,72 @@ const goodsFromServer: string[] = [
...
+  }
+
+  return (
+    <div className="App">
+      {isStarted
+        ? (
+          <>
+            <button type="button" onClick={() => setSortBy(SortType.ALPHA)}>
+              Sort alphabetically
+            </button>
```
jstmpelowycz, line 58:
```suggestion
            <button
              type="button"
              onClick={() => setSortBy(SortType.ALPHA)}
            >
              Sort alphabetically
            </button>
```

jstmpelowycz: suggestion: when passing callbacks to `onClick` and so on, I'd rather create separate function-handlers

src/App.tsx
```diff
@@ -15,33 +15,72 @@ const goodsFromServer: string[] = [
...
+      case SortType.ALPHA:
+        return g1.localeCompare(g2);
+
+      case SortType.LENGTH:
+        return g1.length - g2.length;
+
+      default:
+        return 0;
+    }
+  });
```
jstmpelowycz, line 45: I would suggest one more option you could have used: just create handlers and pass them to buttons, and when those button are clicked – perform the action (i.e. sort, reset, show – and so on)

why: every time your component gets re-rendered – you perform this action, is this okay?

jstmpelowycz (approved):

PR #230, opened 2022-07-18
Solution

yanaost:
[DEMO LINK](https://yanaost.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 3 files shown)


Bohdan-Pyschchyk (approved):

src/App.tsx
```diff
@@ -15,33 +15,104 @@ const goodsFromServer: string[] = [
...
+  if (isReversed === true) {
+    visibleGoods = visibleGoods.reverse();
+  }
+
+  return (
+    visibleGoods.map(item => (
+      <li className="Goods__item" key={item}>{item}</li>
+    ))
+  );
+}
```
jstmpelowycz, line 50: bad idea
the purpose of this function is to return processed data, but you actually make it a component returning you a JSX element, is this OK?
just make it return arrays, and then render those arrays separately

jstmpelowycz (changes requested):

mykola-koval (approved):

PR #231, opened 2022-07-18
goodsHooked

Samaev:
[DEMO LINK](https://Samaev.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
+  const [goodsLength, setGoodsLength] = useState('1');
+  const visibleGoods = goods.filter(good => good.length >= Number(goodsLength));
+
+  if (isSortByLength) {
+    visibleGoods.sort((good1, good2) => good1.length - good2.length);
+  }
+
+  if (isSortByAlphabet) {
+    visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+    }
```
LesyaSvk, line 35: Looks like switch case

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
+        className='title message-header'>
+          Control of goods starts from pushing this
+      </p>
+      <button
+        type="button"
+        className='button is-large is-fullwidth'
+        onClick={() =>setVisio(true)}
+      >
+        Start
     </button>
```
LesyaSvk, line 62:
```suggestion
    {!isVisible && (
      <div>
        <p
          className='title message-header'>
            Control of goods starts from pushing this
        </p>
        <button
          type="button"
          className='button is-large is-fullwidth'
          onClick={() =>setVisio(true)}
        >
          Start
      </button>
```
something went wrong

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
+    </div>
+    }
```
LesyaSvk, line 64:
```suggestion
      </div>
    )}
```

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
 
-    <button type="button">
+{isVisible && <div>
+    <button
+      type="button"
```
LesyaSvk, line 68: here also

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
+        <option value="5">5</option>
+        <option value="6">6</option>
+        <option value="7">7</option>
+        <option value="8">8</option>
+        <option value="9">9</option>
+        <option value="10">10</option>
+      </select>
+    </div>
+
+    </div>}
```
LesyaSvk, line 132: and here

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
+        <li
+          className="Goods__item"
+          key={good}>{good}
+        </li>
+      ))}
     </ul>
+
+    <div>
+      <p className='message-body'>You can choose items to start length from value:</p>
+      <select name="select" value={goodsLength} onChange={({ target }) => setGoodsLength(target.value)}>
```
LesyaSvk, line 118: what about formatting ?

src/App.tsx
```diff
@@ -15,33 +15,121 @@ const goodsFromServer: string[] = [
...
-      <li className="Goods__item">Dumplings</li>
-      <li className="Goods__item">Carrot</li>
-      <li className="Goods__item">Eggs</li>
-      <li className="Goods__item">...</li>
+    <ul className="Goods ">
+      {visibleGoods.map(good =>(
+        <li
+          className="Goods__item"
+          key={good}>{good}
+        </li>
```
LesyaSvk, line 112:
```suggestion
          key={good}
        >
          {good}
        </li>
```

LesyaSvk (changes requested):

src/App.tsx
Samaev: thanks, replaced

src/App.tsx
Samaev: added braces

src/App.tsx
Samaev: here also

src/App.tsx
Samaev: image


I don't understand, but on my screen it looks OK

LesyaSvk (approved):
Good job 🔥

src/App.tsx
```diff
@@ -15,33 +15,140 @@ const goodsFromServer: string[] = [
...
+          </button>
+
+          <button
+            type="button"
+            className="button is-dark"
+            onClick={() => {
+              sortByLength(true);
+              sortAlphabetically(false);
+              setReversed(false);
+            }}
```
jstmpelowycz, line 94: this should not be written in place, you'd rather create a separate function-handler, and then just pass it right here

src/App.tsx
```diff
@@ -15,33 +15,140 @@ const goodsFromServer: string[] = [
...
+            type="button"
+            className="button is-large is-fullwidth"
+            onClick={() => setVisio(true)}
+          >
+            Start
+          </button>
+        </div>
+      )}
+
+      {isVisible && (
```
jstmpelowycz, line 73: you'd rather use ternary here

src/App.tsx
```diff
@@ -15,33 +15,140 @@ const goodsFromServer: string[] = [
...
+      visibleGoods.sort((good1, good2) => good1.length - good2.length);
+      break;
+
+    case isSortByAlphabet:
+      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+      break;
+
+    default:
+      break;
+  }
```
jstmpelowycz, line 40: this solution is weak
I would suggest one more option you could have used: just create handlers and pass them to buttons, and when those button are clicked – perform the action (i.e. sort, reset, show – and so on)

jstmpelowycz (approved):

PR #236, opened 2022-07-18
Solution created

MariiaNH:
[DEMO LINK](https://mariianh.github.io/react_list-of-goods/)


KrisMur (changes requested):
Nice work) 
But have few comments)

PR #248, opened 2022-07-18
list of goods hooks first commit

artemfurdela:
[DEMO LINK](https://artemfurdela.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,33 +15,98 @@ const goodsFromServer: string[] = [
...
+
+  switch (true) {
+    case sortTypeInFuntion === SortType.ALPABET:
+      visibleGoods.sort();
+      break;
+    case sortTypeInFuntion === SortType.LENGTH:
+      visibleGoods.sort((g1, g2) => g1.length - g2.length);
+      break;
+    default:
+      break;
```
valerii-derkach, line 39:
```suggestion
  switch (sortTypeInFuntion) {
    case  SortType.ALPABET:
      visibleGoods.sort();
      break;
    case SortType.LENGTH:
      visibleGoods.sort((g1, g2) => g1.length - g2.length);
      break;
    default:
      break;
```

valerii-derkach: don't repeat the same code, use switch case inside sort method

valerii-derkach (changes requested):

PR #255, opened 2022-07-26
first commit

pavel-gutsal:
https://pavel-gutsal.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,33 +15,107 @@ const goodsFromServer: string[] = [
...
+          type="button"
+          onClick={() => {
+            setStart(false);
+          }}
+        >
+          Start
+        </button>
+      ) : (
+        <>
+          <button
```
alexandra-protyanova, line 68: create separate component for list & buttons

alexandra-protyanova (changes requested):

pavel-gutsal:
sec com

src/App.tsx
```diff
@@ -15,33 +17,63 @@ const goodsFromServer: string[] = [
...
+  const sortHandler = () => {
+    arrayCopy.sort((a, b) => {
+      switch (sortBy) {
+        case 'alphabet':
+          if (!reverse) {
+            return a.localeCompare(b);
+          }
+
+          return b.localeCompare(a);
+          break;
```
valerii-derkach, line 36: no need to use break after return, and better to use ternary instead of `if `statement

src/App.tsx
```diff
@@ -15,33 +17,63 @@ const goodsFromServer: string[] = [
...
+        case 'length':
+          if (!reverse) {
+            return a.length - b.length;
+          }
+
+          return b.length - a.length;
+          break;
+        default:
+          return 0;
+          break;
```
valerii-derkach, line 46: no need to use break after return, and better to use ternary instead of `if `statement

valerii-derkach (changes requested):

pavel-gutsal:
if statement changed for conditional (trinary) operator

marina-tilniak (approved):

PR #260, opened 2022-08-01
looks good

Havrysh:


Havrysh:
https://havrysh.github.io/react_list-of-goods/

src/App.tsx
```diff
@@ -15,33 +16,68 @@ const goodsFromServer: string[] = [
...
+            </button>
+
+            <button type="button" onClick={() => setOrder(true)}>
+              Reset
+            </button>
+
+            {/* Can't sort or reverse the list items after I pressed reset button. Maybe it should be this way if no - let me know please */}
+            {!initialOrder
+              ? <GoodsList goods={visibleGoods} />
+              : <GoodsList goods={goodsFromServer} />}
```
ArturErmolenko, line 76: nope it shouldn't be this way, need to fix this

ArturErmolenko (changes requested):

src/App.tsx
Havrysh: Share the idea

ArturErmolenko: just create handlers and pass them to buttons, and when those button are clicked – perform the action (i.e. sort, reset, show – and so on)

ArturErmolenko (approved):

PR #261, opened 2022-08-03
Solution

BudnikOleksii:
- [DEMO LINK](https://BudnikOleksii.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,7 @@
-import React from 'react';
+import { FC, useState } from 'react';
+import { v4 as uuidv4 } from 'uuid';
 import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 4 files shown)


fledith (approved):
👍

src/App.tsx
```diff
@@ -15,33 +15,118 @@ const goodsFromServer: string[] = [
...
+        default:
+          return 0;
+      }
+    });
+  }
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+};
```
solaryasha, line 57: You should use `useMemo` hook

solaryasha (changes requested):

src/App.tsx
solaryasha: image

solaryasha (changes requested):

src/App.tsx
BudnikOleksii: I will google about this hook, but if I understand correct about useMemo, I call this function only when one of [isReversed, sortType] states changes.
```
const visibleGoods = useMemo(() => (
    getReorderedGoods(preparedGoods, sortType, isReversed)
  ), [isReversed, sortType]);
```

solaryasha: Okay, I see. You moved function out of the component.

solaryasha (approved):

PR #269, opened 2022-08-09
implemented list of goods with hooks

nikita-biriukov:
[DEMO LINK](https://nikita-biriukov.github.io/react_list-of-goods/)

PR #273, opened 2022-08-10
implemented list of goods using hooks

denys-danyliuk:
[DEMO LINK](https://denys-danyliuk.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const goodsFromServer: string[] = [
+
+const goodsFromServer = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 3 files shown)


nikita-biriukov (changes requested):
good job, but [click](https://github.com/mate-academy/react_list-of-goods-hooks/pull/269#pullrequestreview-1068210594)

src/App.tsx
```diff
@@ -15,33 +17,140 @@ const goodsFromServer: string[] = [
...
+      case SortType.ALPABET:
+        return a.localeCompare(b);
+
+      case SortType.LENGTH:
+        return a.length - b.length;
+
+      default:
+        return 0;
+    }
+  });
```
ivan-kadykalo, line 44: This code will work faster because your sorting will not start in case of None of the default, on other hand in your solution it will look at each element and do nothing...

```suggestion
    switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));

      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);

			case SortType.NONE:
      default:
        break;
    }
```

ivan-kadykalo (changes requested):

Ihor-Karpyn (approved):
Good job

PR #275, opened 2022-08-10
Solution

dashasm:
[DEMO LINK](https://dashasm.github.io/react_list-of-goods/)


Ihor-Karpyn (approved):
Good job, but let's add some styles in the next task

PR #277, opened 2022-08-10
Solution

TwoAndMore:
[DEMO LINK](https://TwoAndMore.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+
+import { v4 as uuidv4 } from 'uuid';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const goodsFromServer: string[] = [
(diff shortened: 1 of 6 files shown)


Kras4iy (approved):
👍

src/App.tsx
```diff
@@ -15,33 +17,125 @@ const goodsFromServer: string[] = [
...
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return prev.localeCompare(curr);
+      case SortType.LENGTH:
+        return prev.length - curr.length;
+      case SortType.NONE:
+      default:
+        return 0;
+    }
+  });
```
ivan-kadykalo, line 45: This code will work faster because your sorting will not start in case of None or the default, on other hand in your solution it will look at each element and do nothing...
```suggestion
switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));
      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);
      case SortType.NONE:
      default:
        break;
    }
```

src/App.tsx
```diff
@@ -15,33 +17,125 @@ const goodsFromServer: string[] = [
...
+        <button
+          type="button"
+          className="button is-link"
+          onClick={() => setStart(true)}
+        >
+          Start
+        </button>
+      )}
+
+      {isStarted && (
```
ivan-kadykalo, line 81:
```suggestion
      {!isStarted ? (
        <button
          type="button"
          className="button is-link"
          onClick={() => setStart(true)}
        >
          Start
        </button>
      )
      : (
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
🔥

PR #278, opened 2022-08-11
added list of goods

silchyn:
[DEMO](https://silchyn.github.io/react_list-of-goods-hooks/)


LordAsmodey (approved):
Everything is fine, but styles are needed.

Ihor-Karpyn (approved):

And let's add some style in the next task

PR #279, opened 2022-08-11
Solution

LeonidPustovit:
[DEMO LINK](https://LeonidPustovit.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import './App.css';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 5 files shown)


LeonidPustovit:
I am not sure if we should create new functions, or write in "onClick", So, I left functions, as deleting them is much faster than creating :)

src/App.tsx
```diff
@@ -15,33 +16,160 @@ const goodsFromServer: string[] = [
...
+                <option value="1">1</option>
+                <option value="2">2</option>
+                <option value="3">3</option>
+                <option value="4">4</option>
+                <option value="5">5</option>
+                <option value="6">6</option>
+                <option value="7">7</option>
+                <option value="8">8</option>
+                <option value="9">9</option>
+                <option value="10">10</option>
```
Ihor-Karpyn, line 145: map

src/App.tsx
```diff
@@ -15,33 +16,160 @@ const goodsFromServer: string[] = [
...
+              </select>
+            </div>
+
+            <ul className="Goods">
+              {getReorderedGoods(
+                goodsFromServer,
+                sortType,
+                isReversed,
+                minLength,
```
Ihor-Karpyn, line 154: move to const

Ihor-Karpyn (changes requested):

Something simple like setState u can do inside jsx, but if u create a function it's no mistake

src/App.tsx
```diff
@@ -15,33 +16,163 @@ const goodsFromServer: string[] = [
...
+}
+
+function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+  minLength: number,
+) {
+  // Not to mutate the original array
+  const visibleGoods = [...goods].filter(good => good.length > minLength);
```
solaryasha, line 32: Method `filter` doesn't mutate the array, you don't need to create copy here

src/App.tsx
```diff
@@ -15,33 +16,163 @@ const goodsFromServer: string[] = [
...
+                    </option>
+                  );
+                })}
+              </select>
+            </div>
+
+            <ul className="Goods">
+              {preparedGoods.map(good => {
+                return (
+                  <li className="Goods__item">{good}</li>
```
solaryasha, line 160: You should `key` prop if you rendering the list

solaryasha (changes requested):

solaryasha (approved):

PR #280, opened 2022-08-11
made a list of goods with FC

Oleksandr-Pustovit:
[DEMO LINK](https://Oleksandr-Pustovit.github.io/react_list-of-goods/)


Ihor-Karpyn (approved):

PR #292, opened 2022-08-11
Develop

opanasiukvlad:
[DEMO]{https://opanasiukvlad.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,7 +63,6 @@ export const App: React.FC = () => {
   const handleReverse = () => (setIsReversed(value => !value));
 
   const handleReset = () => {
-    setIsStarted(false);
     setIsReversed(false);
     setSortType(SortType.NONE);
   };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,33 +18,126 @@ const goodsFromServer: string[] = [
...
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+
+      case SortType.LENGTH:
+        return a.length - b.length;
+
+      default:
+        return 0;
+    }
+  });
```
ivan-kadykalo, line 46: Your sort will start every time and will check each element and do nothing

```suggestion
    switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));
      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);
      case SortType.NONE:
      default:
        break;
    }
```

ivan-kadykalo (changes requested):
Such a big field around your button but the clicked area is so small, firstly I thought all this area is a button and it doesn't work. You need to fix this UI-UX.
image

src/App.tsx
opanasiukvlad: Thank you, fixed!

ivan-kadykalo (approved):
🔥

PR #297, opened 2022-08-11
Solution

HennadiiKupr:
[DEMO LINK](https://HennadiiKupr.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,8 @@
-import React from 'react';
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import React, { useState } from 'react';
 import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const goodsFromServer: string[] = [
+const goodsFromServer = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


HennadiiKupr:
I will add CSS today or tomorrow

tetyanabukoros (approved):
Cool))

src/App.tsx
```diff
@@ -15,33 +15,103 @@ const goodsFromServer: string[] = [
...
+        return g1.localeCompare(g2);
+      });
+    }
+
+    if (sortType === SortType.LENGTH) {
+      visibleGoods.sort((g1, g2) => {
+        return g1.length - g2.length;
+      });
+    }
+  }
```
ivan-kadykalo, line 43: Avoid nested if's inside if when it is possible. What will happens without your` if(sortType !== SortType.NONE)` ?

ivan-kadykalo (changes requested):
Don't forget to add styles

src/App.tsx
HennadiiKupr: Will be better to use switch?

ivan-kadykalo: You can, but it is optional in your case, you have only 2 conditions.
So if you will delete 31 line, nothing will change, this is useless line of code.

src/App.tsx
```diff
@@ -15,33 +16,131 @@ const goodsFromServer: string[] = [
...
+}
+
+export const App: React.FC = () => {
+  const [isStarted, setIsStarted] = useState(false);
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const start = () => setIsStarted(true);
+  const reverse = () => setIsReversed(!isReversed);
+  const alphabet = () => setSortType(SortType.ALPABET);
```
VETALEK, line 58: functions should be named as actions they perform 😉 
```suggestion
  const sortByAlphabet = () => setSortType(SortType.ALPABET);
```

VETALEK: variable `alphabet` looks like you just keep a string `'abcd...xyz'`

VETALEK (changes requested):
GJ 👍 
commented

VETALEK (approved):
🔥

ivan-kadykalo (approved):

PR #324, opened 2022-09-06
Develop

Githi54:
[DEMO LINK](https://githi54.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +15,116 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(
+            goodsFromServer,
+            ({
+              sortType,
+              isReversed,
+            }),
+          ).map(good => (
+            <li key={good} data-cy="Good">{good}</li>
```
Githi54, line 126:
```suggestion
            <li key={good} data-cy="Good">
              {good}
            </li>
```

Githi54, line 126:
```suggestion
            <li key={good} data-cy="Good">
              {good}
            </li>
```

aleksandra-prognimak (changes requested):
Read comment

aleksandra-prognimak (approved):
Wonderful🍬

src/App.tsx
```diff
@@ -15,47 +15,119 @@ export const goodsFromServer = [
...
```
dellachaise, line 129: Better create a separate component `GoodsList` for example and render list there.

src/App.tsx
```diff
@@ -15,47 +15,119 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={(
+            sortType === SortType.LENGTH
+              ? 'button is-success'
+              : 'button is-success is-light'
+          )}
```
dellachaise, line 91: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of ternary operator.
The ternary operator is very limited in its functionality to be used in classNames. (this way usually did in real projects and would be great to work with it now).

dellachaise (changes requested):
Great start 👍 

For now, sort works partly incorrect.
When I choose `sort alphabetically then `reverse`(everything great) but if then I choose `sort alphabetically` -  nothing changed(products didn't stay to right alphabetical order).

The same with `sort by length` it doesn't work correctly after reverse.

src/App.tsx
```diff
@@ -15,49 +17,124 @@ export const goodsFromServer = [
...
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames(
+            'button',
+            'is-success',
+            {
+              'is-light':
+                sortType !== SortType.LENGTH,
```
dellachaise, line 96: If line length allows - would be better to write this way.
```suggestion
              'is-light': sortType !== SortType.LENGTH,
```

dellachaise (approved):
Well done 👍

PR #332, opened 2022-09-13
First commit

dm-sheremetiev:
[DEMO LINK](https://dm-sheremetiev.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import React, { useState } from 'react';
+
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +18,122 @@ export const goodsFromServer = [
...
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+// Use this function in the render to prepare goods
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
```
stanislav-topikha, line 35: why object? and if object why only for two parameters?

stanislav-topikha (changes requested):

src/App.tsx
dm-sheremetiev: It's a function by default in task. It's not my. I can fix it

stanislav-topikha (approved):
GJ 👍🏻

PR #335, opened 2022-09-14
Solution

Jamal-Amoian:
[DEMO LINK](https://jamal-amoian.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames(
+            'button',
+            'is-info',
+            { 'is-light': sortType !== SortType.ALPABET },
+          )}
+          onClick={() => setSortType(SortType.ALPABET)}
```
VolodymyrChuyko, line 79: It would be better to implement a handler instead of an inline function. 
Here some [info](https://www.anycodings.com/1questions/4778360/when-to-use-inline-function-on-button-onclick-event-javascriptreactjs)

src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames(
+            'button',
+            'is-success',
+            { 'is-light': sortType !== SortType.LENGTH },
+          )}
+          onClick={() => setSortType(SortType.LENGTH)}
```
VolodymyrChuyko, line 91: Better to use a handler

src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
+            'is-warning',
+            { 'is-light': isReversed !== true },
+          )}
+          onClick={() => {
+            if (isReversed === false) {
+              setIsReversed(true);
+            } else {
+              setIsReversed(false);
+            }
+          }}
```
VolodymyrChuyko, line 109: You can just toggle isReversed also it would be better to use a handler, not an inline function.

VolodymyrChuyko (changes requested):
Good job! Just small issues.

src/App.tsx
Jamal-Amoian: Could you explain how I need toggle "isReversed"?

Jamal-Amoian: I already understood

VolodymyrChuyko (approved):
Well done!

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+  };
+
+  const handleSortByAlphabet = () => {
+    setSortType(SortType.ALPABET);
+  };
+
+  const handleReverse = () => {
+    setIsReversed(value => (
+      !value
+    ));
```
denys-danyliuk, line 68:
```suggestion
    setIsReversed(isReversed => (
      !isReversed
    ));
```

denys-danyliuk (approved):

PR #343, opened 2022-09-16
Solution

Anastasiia-Khmelovska:
[DEMO LINK](https://Anastasiia-Khmelovska.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames(
+            'button is-success',
+            { 'is-light': sortType !== SortType.LENGTH },
+          )}
+          onClick={() => setSortType(SortType.LENGTH)}
```
NataliyaAn, line 90: In this case, you could create the const handleSortLength just like handleReverse.

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames(
+            'button is-info',
+            { 'is-light': sortType !== SortType.ALPABET },
+          )}
+          onClick={() => setSortType(SortType.ALPABET)}
```
NataliyaAn, line 79: In this case, you could create the const handleSortAlphabet just like handleReverse.

NataliyaAn (approved):
Good job!

src/App.tsx
dellachaise: Agreed, need to create separate method instead of creating inline event handler.

src/App.tsx
dellachaise: Agreed, need to create separate method instead of creating inline event handler.

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goods.map((good) => (
+          <li data-cy="Good" key={good}>
+            {good}
+          </li>
+        ))}
```
dellachaise, line 121: Better create a separate component to render GoodsList.

dellachaise (changes requested):
Good job 👍 

Just add changes according to comments 👇

src/App.tsx
```diff
@@ -15,48 +16,117 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPABET,
```
AndriiRut, line 21:
```suggestion
  ALPHABET,
```

AndriiRut (approved):
🚀

PR #352, opened 2022-09-19
Solution

ostUp:
[DEMO LINK](https://ostUp.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,48 +17,113 @@ export const goodsFromServer = [
...
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort();
+  }
```
max-kravchenko, line 39: I guess here it is preferable to use localeCompare here (in case our list includes some other starting letters in future)

p.s. And in general switch/case construct is considered to be better if we have some strict sorting options

max-kravchenko (approved):
Good job!

p.s. Only minor comments related to getReorderedGoods function

AndriiRut (approved):
Good job!

PR #357, opened 2022-09-19
commit message

11Olya112:


src/App.tsx
```diff
@@ -15,48 +17,124 @@ export const goodsFromServer = [
...
+  ALPABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+const getReorderedGoods = function (
```
Bohdan-Maksymiuk, line 31: use arrow function

src/App.tsx
```diff
@@ -15,48 +17,124 @@ export const goodsFromServer = [
...
+
+  const handleSortByAlpabet = () => {
+    setSortType(SortType.ALPABET);
+  };
+
+  const handleSortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const NONE = () => {
```
Bohdan-Maksymiuk, line 72: bad name, better `resetSortType` or smth like that

src/App.tsx
```diff
@@ -15,48 +17,124 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {condition
```
Bohdan-Maksymiuk, line 120: it is not obvious from name what is the condition
use `shouldRenderResetButton` or smth like that

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #385, opened 2022-09-21
Solution

Black3942:


alexandra-protyanova (approved):
I think it is better to create separate component for goods

PR #398, opened 2022-09-30
Solution

ElinaBahirova:
https://elinabahirova.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
+      return good1.localeCompare(good2);
+    }
+
+    if (sortType === SortType.LENGTH) {
+      return good1.length - good2.length;
+    }
+
+    return 1;
+  });
+
```
SanyaBratashchuk, line 45: use switch case

src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
+                className="button is-danger is-light"
+                onClick={() => {
+                  reverse(false);
+                  setSortType(SortType.NONE);
+                }}
+              >
+                Reset
+              </button>
+            )
+            : ('')}
```
SanyaBratashchuk, line 111:
```suggestion
          {isReversed || sortType !== SortType.NONE
            && (
              <button
                type="button"
                className="button is-danger is-light"
                onClick={() => {
                  reverse(false);
                  setSortType(SortType.NONE);
                }}
              >
                Reset
              </button>
            )
            }
```

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,106 @@ export const goodsFromServer = [
...
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames(
+            'button',
+            'is-info',
+            { 'is-light': sortType !== SortType.ALPABET },
+          )}
+          onClick={() => setSortType(SortType.ALPABET)}
```
liliia-livandovskaya, line 70: it's not a mistake, but it would be better to add a handler on each click

liliia-livandovskaya (approved):
GJ 🔥

PR #401, opened 2022-10-05
done

ghost:
[DEMO LINK](https://kamal-tufekcic.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
+import './App.scss';
 import React from 'react';
 import 'bulma/css/bulma.css';
-import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -21,7 +22,7 @@ enum SortType {
   LENGTH,
 }
 
-export function getReorderedGoods(
+function getReorderedGoods(
   goods: string[],
   sortType: SortType,
   isReversed: boolean,
@@ -39,7 +40,9 @@ export function getReorderedGoods(
     }
   });
 
-  return isReversed ? visibleGoods.reverse() : visibleGoods;
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
 }
 
 export const App: React.FC = () => {
@@ -63,7 +66,12 @@ export const App: React.FC = () => {
         <button
           onClick={sortByName}
           type="button"
-          className={`button is-outlined is-success is-rounded ${sortType !== SortType.ALPHABET && 'is-light'}`}
+          className={classNames({
+            'button is-outlined is-success is-rounded':
+              sortType === SortType.ALPHABET,
+            'button is-outlined is-success is-rounded is-light':
+              sortType !== SortType.ALPHABET,
+          })}
         >
           Sort alphabetically
         </button>
@@ -71,7 +79,12 @@ export const App: React.FC = () => {
         <button
           onClick={sortByLength}
           type="button"
-          className={`button is-rounded is-danger is-outlined ${sortType !== SortType.LENGTH && 'is-light'}`}
+          className={classNames({
+            'button is-outlined is-danger is-rounded':
+              sortType === SortType.LENGTH,
+            'button is-outlined is-danger is-rounded is-light':
+              sortType !== SortType.LENGTH,
+          })}
         >
           Sort by length
         </button>
@@ -79,13 +92,16 @@ export const App: React.FC = () => {
         <button
           onClick={reverse}
           type="button"
-          className={`button is-rounded is-black is-outlined ${!isReversed && 'is-light'}`}
+          className={classNames({
+            'button is-outlined is-black is-rounded': isReversed,
+            'button is-outlined is-black is-rounded is-light': !isReversed,
+          })}
         >
           Reverse
         </button>
 
         {(sortType || isReversed)
-          ? (
+          && (
             <button
               type="button"
               className="button is-rounded is-ghost is-outlined"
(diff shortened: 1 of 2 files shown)


Quelaan1 (approved):

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+export function getReorderedGoods(
```
ihor-jpeg, line 24: that moment when you export a function, but your entire app is one file 😆

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
+      case SortType.ALPHABET:
+        return g1.localeCompare(g2);
+      case SortType.LENGTH:
+        return g1.length - g2.length;
+      default:
+        return 0;
+    }
+  });
+
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
```
ihor-jpeg, line 42:
```suggestion
  return isReversed
    ? visibleGoods.reverse()
    : visibleGoods;
```

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
-      <div className="buttons">
+      <div className="buttons is-flex is-centered">
         <button
+          onClick={sortByName}
           type="button"
-          className="button is-info is-light"
+          className={`button is-outlined is-success is-rounded ${sortType !== SortType.ALPHABET && 'is-light'}`}
```
ihor-jpeg, line 66: God noooo
let's use `classnames` npm package

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
         >
           Sort alphabetically
         </button>
 
         <button
+          onClick={sortByLength}
           type="button"
-          className="button is-success is-light"
+          className={`button is-rounded is-danger is-outlined ${sortType !== SortType.LENGTH && 'is-light'}`}
```
ihor-jpeg, line 74: classnames

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
         >
           Sort by length
         </button>
 
         <button
+          onClick={reverse}
           type="button"
-          className="button is-warning is-light"
+          className={`button is-rounded is-black is-outlined ${!isReversed && 'is-light'}`}
```
ihor-jpeg, line 82: classnames

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType || isReversed)
```
ihor-jpeg, line 87: if you are choosing between showing an element or not based on a condition, a cleaner way is to use logical &&

src/App.tsx
```diff
@@ -15,49 +15,105 @@ export const goodsFromServer = [
...
+              onClick={reset}
+            >
+              Reset
+            </button>
+          )
+          : ''}
+      </div>
+      <div className="is-flex is-justify-content-center">
+        <div className="has-text-centered">
+          {(getReorderedGoods(goodsFromServer, sortType, isReversed))
```
ihor-jpeg, line 101: you will call this bad boy every time a render happens
you can create a variable with these goods before the return, memoize it (useMemo with a dependency array, so it will be reordered only when `goodsFromServer, sortType, isReversed` change) and then use it inside of the return

it's not too big of a problem here, but if this value was not just inside div, but passed to another component, it would make an impact, because that component (and possibly its children would be rerendered every single time, regardless of whether `goodsFromServer, sortType, isReversed` changed)

ihor-jpeg (changes requested):

src/App.tsx
kamal-tufekcic: Meanwhile, I was intentionally avoiding using it :|

src/App.tsx
kamal-tufekcic: JS does have garbage collection, but I don't :|

src/App.tsx
kamal-tufekcic: I'll need to do some research before implementing that, the rest was fixed

src/App.tsx
```diff
@@ -63,38 +66,50 @@ export const App: React.FC = () => {
         <button
           onClick={sortByName}
           type="button"
-          className={`button is-outlined is-success is-rounded ${sortType !== SortType.ALPHABET && 'is-light'}`}
+          className={classNames({
+            'button is-outlined is-success is-rounded':
+              sortType === SortType.ALPHABET,
+            'button is-outlined is-success is-rounded is-light':
+              sortType !== SortType.ALPHABET,
+          })}
```
ihor-jpeg, line 74: before you had this 
`className={`button is-outlined is-success is-rounded ${sortType !== SortType.ALPHABET && 'is-light'}`}`

from what i can see, the only optional class is 'is-light'

classnames allow you to write it cleaner

import cn from 'classnames';

cn(
  'button',
  'is-outlined',
  'is-success',
  'is-rounded',
  { 'is-light': sortType === SortType.ALPHABET },
)

what you have wrote in the previous commit:
`button is-outlined is-success is-rounded ${sortType !== SortType.ALPHABET && 'is-light'}`
will produce (`button is-outlined is-success is-rounded` + 'is-light' or +'false')

which is not correct (you have 'false' class, yikes)
what you have written in this commit is okay, but you dont need two long strings

ihor-jpeg (approved):

PR #402, opened 2022-10-05
Develop

Quelaan1:
- [DEMO LINK](https://Quelaan1.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -49,13 +49,13 @@ export const App: React.FC = () => {
   const [isReversed, setReversed] = useState(false);
   const [sortType, setSortType] = useState(SortType.NONE);
 
-  const alphabet = () => setSortType(() => SortType.ALPHABET);
+  const handleSoryByAlphabet = () => setSortType(() => SortType.ALPHABET);
 
-  const length = () => setSortType(() => SortType.LENGTH);
+  const handleSortByLength = () => setSortType(() => SortType.LENGTH);
 
-  const reverse = () => setReversed(() => !isReversed);
+  const handleSortReverse = () => setReversed(() => !isReversed);
 
-  const reset = () => {
+  const handleReset = () => {
     setReversed(false);
     setSortType(SortType.NONE);
   };
@@ -73,7 +73,7 @@ export const App: React.FC = () => {
               'is-light': sortType !== SortType.ALPHABET,
             },
           )}
-          onClick={alphabet}
+          onClick={handleSoryByAlphabet}
         >
           Sort alphabetically
         </button>
@@ -86,7 +86,7 @@ export const App: React.FC = () => {
               'button is-info is-light': sortType !== SortType.LENGTH,
             },
           )}
-          onClick={length}
+          onClick={handleSortByLength}
         >
           Sort by length
         </button>
@@ -99,7 +99,7 @@ export const App: React.FC = () => {
               'button is-info is-light': isReversed === false,
             },
           )}
-          onClick={reverse}
+          onClick={handleSortReverse}
         >
           Reverse
         </button>
@@ -108,7 +108,7 @@ export const App: React.FC = () => {
           <button
             type="button"
             className="button is-danger is-light"
-            onClick={reset}
+            onClick={handleReset}
           >
             Reset
           </button>


kamal-tufekcic (approved):

src/App.tsx
```diff
@@ -15,49 +16,121 @@ export const goodsFromServer = [
...
+        return g1.localeCompare(g2);
+      });
+    }
+
+    if (sortType === SortType.LENGTH) {
+      visibleGoods.sort((g1, g2) => {
+        return g1.length - g2.length;
+      });
+    }
+  }
```
ihor-jpeg, line 44: use switch

src/App.tsx
```diff
@@ -15,49 +16,121 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const alphabet = () => setSortType(SortType.ALPHABET);
+
+  const length = () => setSortType(SortType.LENGTH);
+
+  const reverse = () => setReversed(!isReversed);
```
ihor-jpeg, line 61: if you set state based on previous state, which you do, use functional set state

src/App.tsx
```diff
@@ -15,49 +16,121 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames({
+            'button is-info': sortType === SortType.ALPHABET,
+            'button is-info is-light': sortType !== SortType.ALPHABET,
+          })}
```
ihor-jpeg, line 78:
```suggestion
          className={classNames(
            'button is-info',
            {
              'is-light': sortType !== SortType.ALPHABET
            },
          })}
```

src/App.tsx
```diff
@@ -15,49 +16,121 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {sortType !== SortType.NONE || isReversed !== false ? (
```
ihor-jpeg, line 106: why not logical &&?

ihor-jpeg (changes requested):

src/App.tsx
```diff
@@ -15,49 +16,122 @@ export const goodsFromServer = [
...
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const alphabet = () => setSortType(() => SortType.ALPHABET);
+
+  const length = () => setSortType(() => SortType.LENGTH);
```
ihor-jpeg, line 54: no need to do it here. here you do not use the previous state value

src/App.tsx
```diff
@@ -15,49 +16,122 @@ export const goodsFromServer = [
...
+
+  const reverse = () => setReversed(() => !isReversed);
```
ihor-jpeg, line 56: but here you do need the previous to set the next value
the only thing is that the functional set state accepts the prev state as an argument, you need to use it, otherwise right now you are still not using the prev state value properly

```suggestion
  const reverse = () => setReversed((isReversed) => !isReversed);
```

ihor-jpeg (changes requested):

src/App.tsx
ihor-jpeg, line 56: also, choose better names

if it's a handler, maybe
handleReset
handleSortByLength
handleSoryByAlphabet

ihor-jpeg (changes requested):

src/App.tsx
```diff
@@ -15,49 +16,122 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const handleSoryByAlphabet = () => setSortType(() => SortType.ALPHABET);
+
+  const handleSortByLength = () => setSortType(() => SortType.LENGTH);
+
+  const handleSortReverse = () => setReversed(() => !isReversed);
```
ihor-jpeg, line 56:
```suggestion
  const handleSoryByAlphabet = () => setSortType(SortType.ALPHABET);

  const handleSortByLength = () => setSortType(SortType.LENGTH);

  const handleSortReverse = () => setReversed((state) => !state);
```

ihor-jpeg (approved):

PR #403, opened 2022-10-05
solution

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,39 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+
+enum SortType {
+  NONE = 0,
+  ALPABET = 1,
+  LENGTH = 2,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  } else if (sortType === SortType.ALPABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,46 +49,72 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, {
+            sortType: sortBy,
+            isReversed,
+          }).map((good: string) => {
+            return (
+              <li key={good} data-cy="Good">{good}</li>
+            );
+          })}
```
silchyn, line 117:
```suggestion
          }).map((good: string) => (
            <li key={good} data-cy="Good">{good}</li>
          ))}
```

silchyn (changes requested):

src/App.tsx
```diff
@@ -16,46 +49,72 @@ export const goodsFromServer = [
...
+          {getReorderedGoods(goodsFromServer, {
+            sortType: sortBy,
+            isReversed,
+          }).map((good: string) => (
+            <li key={good} data-cy="Good">
+              {good}
+            </li>
+          ))}
         </ul>
       </ul>
```
silchyn, line 119: Is this nested 'ul' necessary?

silchyn (changes requested):

src/App.tsx
```diff
@@ -16,47 +49,71 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getReorderedGoods(goodsFromServer, {
+          sortType: sortBy,
+          isReversed,
+        }).map((good: string) => (
```
Andr1yK, line 112: do it outside return

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -1,6 +1,39 @@
...
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  } else if (sortType === SortType.ALPABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
```
Anna-Tkachenko, line 29: if you need to do some logic depending on the value of some variable (sortType here), will be better to use switch / case

src/App.tsx
```diff
@@ -16,47 +49,77 @@ export const goodsFromServer = [
...
-          className="button is-info is-light"
+          className={classNames('button is-info', {
+            'is-light': sortBy !== SortType.ALPABET,
+          })}
+          onClick={() => {
+            setSortBy(SortType.ALPABET);
+          }}
         >
           Sort alphabetically
         </button>
```
Anna-Tkachenko, line 82: some logic for each button is repeated, will be great to find a way to move it to some separate component, that can receive needed props

Anna-Tkachenko (approved):
⭐️

PR #409, opened 2022-10-11
solution

Naveed-Shoukat:
[DEMO LINK](https://naveed-shoukat.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,47 +15,140 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  let visibleGoods = [...goods];
+
+  switch (sortType) {
+    case SortType.ALPABET:
+      visibleGoods = visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods = visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+      //
+  }
+
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [appState, setAppState] = React.useState({
+    sortType: SortType.NONE,
+    isReversed: false,
+  });
+
+  const handelSortAlphbet = () => {
+    setAppState(prevState => {
+      return {
+        ...prevState,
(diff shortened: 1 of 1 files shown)


fatihsamur (approved):
works well

src/App.tsx
```diff
@@ -15,47 +15,140 @@ export const goodsFromServer = [
...
+    case SortType.ALPABET:
+      visibleGoods = visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods = visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+      //
```
ihor-jpeg, line 45: returning the original array here would be a good idea

src/App.tsx
```diff
@@ -15,47 +15,140 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={appState.sortType === SortType.ALPABET
+            ? 'button is-info'
+            : 'button is-info is-light'}
```
ihor-jpeg, line 102: checkout classnames lib

src/App.tsx
```diff
@@ -15,47 +15,140 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(appState.isReversed || appState.sortType !== SortType.NONE) && (
```
ihor-jpeg, line 128: such long conditions are better to put in a varibale and use just one variable here

src/App.tsx
```diff
@@ -15,47 +15,140 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(
+            goodsFromServer,
+            appState,
+          ).map(good => (
```
ihor-jpeg, line 144: create variable. you can also memoize it (useMemo) so you will have a performance increase, wherein the current setup you always run the function getReorderedGoods, even if the order did not change

ihor-jpeg (changes requested):

src/App.tsx
Naveed-Shoukat: applied to all the three buttons.

src/App.tsx
Naveed-Shoukat: Variable created for storing sorted goods and React useMemo was used for optimized performance. Also the App states was NOT modified if the prevState was unchange. So that clicking any sort button more than once will not re-render the App for optimized performacne.

ihor-jpeg (approved):

PR #413, opened 2022-10-17
created list on hooks

AntonVaida:
[DEMO LINK](https://AntonVaida.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,133 @@ export const goodsFromServer = [
...
+
+    case SortType.LENGTH:
+      resultList = visibleGoods.sort((goodA, goodB) => (
+        goodA.length - goodB.length
+      ));
+      break;
+    default:
+      resultList = visibleGoods;
+      break;
+  }
```
illia-sosnytskyi, line 52: use `switch` condition inside `sort` method to avoid repeating this method

OleksiiBlezniuk: agree, but it's not critical

src/App.tsx
```diff
@@ -15,48 +16,133 @@ export const goodsFromServer = [
...
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+  let resultList;
```
OleksiiBlezniuk, line 35: no need to create extra variable, just mutate `visibleGoods`

src/App.tsx
```diff
@@ -15,48 +16,133 @@ export const goodsFromServer = [
...
+  const sortByAlphabet = () => {
+    setSortType(SortType.ALPABET);
+  };
+
+  const sortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const makeReverse = () => {
+    setIsReversed(!isReversed);
```
OleksiiBlezniuk, line 75: don't pass state value to setState function, you can pass function here (like functional setState, but in function returned from hook)

OleksiiBlezniuk (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,127 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+
+  // eslint-disable-next-line react/sort-comp
```
Andr1yK, line 59: do u need this comment?

Andr1yK (approved):

PR #429, opened 2022-10-21
list of goods(hooks) imlemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { v4 as uuid } from 'uuid';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -15,47 +17,105 @@ export const goodsFromServer = [
...
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, toggleReverse] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const reorderedGoods = getReorderedGoods(
```
TheVovchik, line 55: maybe if we return visibleGoods in function let it be here the same? 
think names need to be close one another

src/App.tsx
```diff
@@ -15,47 +17,105 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed) && (
```
TheVovchik, line 100: think we need to move this condition to separate value, it will improve readability

src/App.tsx
```diff
@@ -15,47 +17,105 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortType(SortType.NONE);
+              toggleReverse(false);
```
TheVovchik, line 106: I think with named functions would be better

TheVovchik (approved):
Look on naming, and once more about condition you can find that it's common reject in many tasks

good rework of class component 🔥

src/App.tsx
```diff
@@ -15,47 +17,119 @@ export const goodsFromServer = [
...
+
+      case SortType.LENGTH:
+        return goodA.length - goodB.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  return isReversed
```
naziamoff, line 46: I believe it is much better to separate the logic of reverse, because now we check the condition each render, even if the button was not clicked

src/App.tsx
```diff
@@ -15,47 +17,119 @@ export const goodsFromServer = [
...
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, toggleReverse] = useState(false);
```
naziamoff, line 52: don't think it is critical in this case, but it is better to stick to the useState hook naming conventions - it should be `setIsReversed` in order to make it clear that we using hook

src/App.tsx
```diff
@@ -15,47 +17,119 @@ export const goodsFromServer = [
...
+
+  const sortAlphabetically = () => {
+    setSortType(SortType.ALPABET);
+  };
+
+  const sortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const reverse = () => {
```
naziamoff, line 63: and this function definitely should be called toggleReverse

src/App.tsx
```diff
@@ -15,47 +17,119 @@ export const goodsFromServer = [
...
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames(
+            'button',
+            'is-success',
+            { 'is-light': sortType !== SortType.LENGTH },
+          )}
+          onClick={sortByLength}
```
naziamoff, line 100: 👍

src/App.tsx
naziamoff: Totally agree with @TheVovchik 
What is more, if you need some changes in this condition later - you will need to look for this part of code. With variable it is much more simple

src/App.tsx
```diff
@@ -15,47 +17,119 @@ export const goodsFromServer = [
...
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map(good => (
+            <li key={uuid()} data-cy="Good">{good}</li>
```
naziamoff, line 131: we generate new uuid each time, that means that react rerenders the list each time. The key should be unique, but not random

naziamoff (changes requested):

src/App.tsx
artem-musii: I have moved logic of reverse, but I still check it every render, so maybe I understood you wrong?

src/App.tsx
```diff
@@ -1,61 +1,142 @@
...
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  { name: 'Dumplings', id: uuid() },
```
naziamoff, line 8: great job 👍

naziamoff (approved):
👍

src/App.tsx
naziamoff: you understood me correct, well done. You will get the point as soon as you get acquainted with memoization - that helps to save function without rerendering them each time if not needed.

PR #435, opened 2022-10-21
Implemented list of goods use hooks

jojnajbude:
[DEMO LINK](https://jojnajbude.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
-export const goodsFromServer = [
+type Goods = string[];
+
+export const goodsFromServer: Goods = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +18,104 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  NONE,
+  ALPABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const [isReversed, Reverse] = useState(false);
```
grafit-off, line 28: Maybe it's not critical, but usually, we use the prefix `set` for the function returned by `useState` to change the state. In the case of `Reverse` in the code below it's not really clear to me what function `Reverse` means, so I need to find it. But as I said before, maybe, it's not critical, and it's only my paranoia

grafit-off (approved):
GJ 🔥

src/App.tsx
jojnajbude: thanks, good comment

src/App.tsx
```diff
@@ -15,48 +18,104 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  NONE,
+  ALPABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReverse] = useState(false);
+  const [sortBy, chooseSort] = useState(SortType.NONE);
```
Denchepornyuk, line 29: add type for useState - `sortBy`

src/App.tsx
```diff
@@ -15,48 +18,104 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getReorderedGoods(goodsFromServer, sortBy)
+          .map(good => (
+            <li data-cy="Good" key={good}>{good}</li>
+          ))}
```
Denchepornyuk, line 118: create a separate component for goods list

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #438, opened 2022-10-22
React list of goods (Hooks)

Vladosik1337:
[DEMO LINK](https://Vladosik1337.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


Anna-Borodaienko (approved):

src/App.tsx
```diff
@@ -15,48 +17,101 @@ export const goodsFromServer = [
...
+  sortType: SortType,
+  isReversed: boolean,
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  } else if (sortType === SortType.ALPABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
```
naziamoff, line 37: switch/case will do better here

src/App.tsx
```diff
@@ -15,48 +17,101 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPABET,
+  LENGTH,
+}
+
+export function getReorderedGoods(
```
naziamoff, line 26: it is a bad practice to accept a lot of arguments in the function, it makes your code more complicated and less possible to reuse later

src/App.tsx
```diff
@@ -15,48 +17,101 @@ export const goodsFromServer = [
...
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReverse] = useState(false);
+  const isDefault = sortType !== SortType.NONE || isReversed;
```
naziamoff, line 49: this variable name is lying, your conditions actually describe the changed array, not the default.

naziamoff (changes requested):

src/App.tsx
```diff
@@ -15,48 +17,119 @@ export const goodsFromServer = [
...
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
```
naziamoff, line 33: I think it would be better to separate reversing into different function
Now we are checking the condition of reversing each render, even if the button reverse was not clicked

src/App.tsx
```diff
@@ -15,48 +17,119 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReverse] = useState(false);
```
naziamoff, line 62:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +17,119 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map(good => (
+          <li key={uuidv4()} data-cy="Good">{good}</li>
```
naziamoff, line 131: the key should be unique but not random. Now we rerender the list each time because each time we have new key there

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #439, opened 2022-10-22
List of goods has been created

iam-datsenko:


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <button
+      type="button"
+      className={classNames('button', className)}
+      {...rest}
+    >
+      {children}
+    </button>
+  );
+};
```
iPanda177, line 24: it's cool that you decided to make separate components, but I think it only complicated your code in this case

iPanda177 (approved):

src/components/Button/Button.tsx
iam-datsenko: This is a good practice before real work, so I get used to writing the way I'm supposed to))

PR #445, opened 2022-10-23
implemented a list of goods on the functional component with hooks usage

artem-nasadchuk:
[DEMO LINK](https://artem-nasadchuk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cl from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


anton-iskryk (approved):

src/App.tsx
```diff
@@ -15,47 +17,136 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setReverse] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const visibleGoods = getReorderedGoods(
+    goodsFromServer, { isReversed, sortType },
+  );
+
+  const isOriginal = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 65: so due to this logic the array is original if it is reversed or sorted? Looks like the name of the variable is lying to me :harold

src/App.tsx
```diff
@@ -15,47 +17,136 @@ export const goodsFromServer = [
...
           type="button"
-          className="button is-info is-light"
+          className={cl(
+            'button',
+            'is-info',
+            {
+              'is-light': sortType !== SortType.ALPHABET,
+            },
+          )}
+          onClick={sortByAlphabet}
```
naziamoff, line 96: 👍

src/App.tsx
```diff
@@ -15,47 +17,136 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {isOriginal && (
```
naziamoff, line 129: you should definitely fix this, because now it seems that we render this button only if the array is not changed what makes no sense

src/App.tsx
```diff
@@ -15,47 +17,136 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map(good => (
+            <li
+              data-cy="Good"
+              key={uuidv4()}
```
naziamoff, line 145: we gonna add a new key on each render, that is not what we need. Key should be unique but not random

naziamoff (changes requested):

src/App.tsx
artem-nasadchuk: i can use 'good' as a key here?

src/App.tsx
artem-nasadchuk: agree

src/App.tsx
naziamoff: Ideally the ids should be generated for each good, but in this case indeed.

naziamoff (approved):

PR #453, opened 2022-10-24
commit1

Gintako666:
[DEMO LINK](https://Gintako666.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import className from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type SortType = 'NONE' | 'ALPABET' | 'LENGTH';
+
+export function GetReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+) {
```
stanislav-topikha, line 25: it`s not ok to return jsx like this in separate function, use this function to reorder items and return reordered list, then map it inside of App component

and why do you need export here?

stanislav-topikha (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type SortType = 'NONE' | 'ALPABET' | 'LENGTH';
+
+function GetReorderedGoods(
```
DarkMistyRoom, line 21:
```suggestion
function getReorderedGoods(
```

DarkMistyRoom (changes requested):

stanislav-topikha (approved):

PR #454, opened 2022-10-24
React list of goods implemented with functional component and Hooks

TheVovchik:
**[DEMO LINK](https://TheVovchik.github.io/react_list-of-goods-hooks/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { v4 as uuidv4 } from 'uuid';
+import { Button } from './Components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


artem-musii (approved):
GJ!

src/App.tsx
```diff
@@ -15,47 +18,109 @@ export const goodsFromServer = [
...
+  LENGTH,
+}
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export function getSortedGoods(
+  goods: string[],
+  state: State,
```
naziamoff, line 33: I don't think that passing state as props is a good idea here, because after adding other props to the state - we will pass the args we don't need. 

What I meant with my comment about avoiding too many props in the func is that you should separate functions, but not pass props as objects. Feel free to ask me if you have further questions about this comment

naziamoff (approved):
GJ. I have one comment though

src/App.tsx
TheVovchik: will ask on QnA

PR #457, opened 2022-10-24
Implement list of goods with hooks

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,47 +17,122 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(
+            goodsFromServer,
+            sortType,
+            isReversed,
+          ).map(good => (
```
artem-nasadchuk, line 133: for better readability, it is better to put the result of the function into a variable and here map this variable

src/App.tsx
```diff
@@ -15,47 +17,122 @@ export const goodsFromServer = [
...
+            <li key={uuidv4()} data-cy="Good">{good}</li>
```
artem-nasadchuk, line 134:
```suggestion
            <li
              key={uuidv4()}
              data-cy="Good"
            >
              {good}
            </li>
```

artem-nasadchuk (approved):

src/App.tsx
```diff
@@ -15,47 +17,130 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
+export const App: React.FC<{}> = () => {
+  const [isReversed, setReverse] = useState(false);
```
naziamoff, line 54: Normally we use value/setValue naming in hooks
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +17,130 @@ export const goodsFromServer = [
...
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+  const visibilityResetButton = sortType !== SortType.NONE || isReversed;
```
naziamoff, line 56: boolean variables should start with `should`, `can`, `is` and similar

src/App.tsx
```diff
@@ -15,47 +17,130 @@ export const goodsFromServer = [
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleItems.map(good => (
+            <li
+              key={uuidv4()}
```
naziamoff, line 138: the key should be unique but not random. Now we rerender the list each time because each time we have new key there

naziamoff (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+        return currentGood.length - nextGood.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
naziamoff, line 47: actually it would be great to separate the reverse logic, it will help with code maintenance in the future.
You should avoid accepting more than 2 args in the function if it is possible. That makes your code more complicated

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+}
+
+export const App: React.FC<{}> = () => {
+  const [isReversed, setIsReverse] = useState(false);
```
naziamoff, line 53: I'm not grammar nazi, but
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

naziamoff (approved):
Well done 🔥

PR #458, opened 2022-10-24
React list of goods (hooks)

Yuriy-Komolov:
[DEMO LINK](https://Yuriy-Komolov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import { GoodsList } from './components/GoodsList';
+import { Button } from './components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


igor-1987 (approved):
GJ!

src/App.tsx
```diff
@@ -15,49 +20,109 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const sortByLength = () => setSortType(SortType.LENGTH);
+
+  const sortByAlphabet = () => setSortType(SortType.ALPHABET);
+
+  const reverse = () => setIsReversed((state) => !state);
```
naziamoff, line 68: Better to write it explicitly
```suggestion
  const toggleReverse = () => setIsReversed((state) => !state);
```

src/App.tsx
```diff
@@ -15,49 +20,109 @@ export const goodsFromServer = [
...
-        </button>
+        </Button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed) && (
```
naziamoff, line 115: if you will need to change the condition in the future - you will have to look for this peace of code in order to change it. So it is better to create a separate variable for that

naziamoff (approved):

PR #459, opened 2022-10-24
rewrote app into functional component

poznianski:
[DEMO LINK](https://poznianski.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


MAKSOti (approved):
cool)

src/App.tsx
```diff
@@ -15,47 +16,120 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+// Use this function in the render to prepare goods
```
naziamoff, line 30: you should remove all unnecessary comments

src/App.tsx
```diff
@@ -15,47 +16,120 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  // state = {
+  //   isReversed: false,
+  //   sortType: SortType.NONE,
+  // };
```
naziamoff, line 62: never ever leave commented code

src/App.tsx
```diff
@@ -15,47 +16,120 @@ export const goodsFromServer = [
...
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {
+          (sortType || isReversed)
```
naziamoff, line 113: better to make it a separate variable. Will be easier to maintain this code later

naziamoff (changes requested):
Well done 🔥 
I can't approve it right away though, because leaving comments is more critical than it may seem

Anna-Tkachenko (approved):
⭐️

PR #472, opened 2022-10-26
create react list og doods hooks

MAKSOti:
[DEMO LINK](https://MAKSOTi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { v4 as uuidv4 } from 'uuid';
+import { Button } from './Components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


poznianski (approved):
good

src/App.tsx
```diff
@@ -15,47 +18,109 @@ export const goodsFromServer = [
...
+  LENGTH,
+}
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export function getSortedGoods(
+  goods: string[],
+  state: State,
```
naziamoff, line 33: you should not pass the state inside the function. We should use only those props we need.
What's more, I believe that reverse logic should be separated. 

The point is the function ideally has but one single responsibility, it should not do a lot of stuff.

src/App.tsx
```diff
@@ -15,47 +18,109 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const resetOrder = () => {
+    setIsReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const visibleGoods = getSortedGoods(goodsFromServer,
+    { sortType, isReversed });
```
naziamoff, line 64: well k, what you're passing to the function is not a state actually. Why is it called like that then?

src/App.tsx
```diff
@@ -15,47 +18,109 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map(good => (
+            <li
+              data-cy="Good"
+              key={uuidv4()}
```
naziamoff, line 119: you should not generate random key. That means you gonna rerender a list with each component rendering - you definitely don't want it to be that way.

naziamoff (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #473, opened 2022-10-26
Solution

tatyana22:
[DEMO LINK](https://tatyana22.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: Goods = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,105 @@ export const goodsFromServer = [
...
+    visibleGoods.sort((good1, good2) => {
+      switch (sortBy) {
+        case SortType.ALPABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+        default:
+          return 0;
+      }
+    });
```
Andr1yK, line 46:
```suggestion
    const visibleGoods = [...goods].sort((good1, good2) => {
      switch (sortBy) {
        case SortType.ALPABET:
          return good1.localeCompare(good2);
        case SortType.LENGTH:
          return good1.length - good2.length;
        default:
          return 0;
      }
    });
```

Andr1yK (approved):

PR #477, opened 2022-10-31
solution

pushkalov:
https://<pushkalov.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+      default:
+        return 0;
```
Anastasiia-Svintsova, line 39: It will be easier to read
```suggestion
      case SortType.ALPHABET:
        return good1.localeCompare(good2);
        
      case SortType.LENGTH:
        return good1.length - good2.length;
        
      default:
        return 0;
```

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+
+  const clickAlphabet = () => {
+    setSortType(SortType.ALPHABET);
+  };
+
+  const clickReversed = () => {
+    setReverse(value => (
+      !value
+    ));
+  };
```
Anastasiia-Svintsova, line 62:
```suggestion
  const clickReversed = () => {
    setReverse(!isReversed);
  };
```

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReverse] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const clickAlphabet = () => {
```
Anastasiia-Svintsova, line 54: Let's rename functions to show that it's actions
```suggestion
  const onClickAlphabet = () => {
```

Anastasiia-Svintsova (changes requested):

polosanya (approved):
Well done!

PR #478, opened 2022-11-02
solution

mayuces:
[DEMO LINK](https://myauces.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +15,119 @@ export const goodsFromServer = [
...
+  });
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, sortTypeSetter] = useState(SortType.NONE);
+  const [isReversed, isReversedSetter] = useState(false);
```
hajnaloltyan, line 53: I think it should be named setSortType rather than sortTypeSetter. Same with isReversed

src/App.tsx
```diff
@@ -15,47 +15,119 @@ export const goodsFromServer = [
...
+
+  const reverseHandler = () => {
+    isReversedSetter(prevState => !prevState);
```
hajnaloltyan, line 56: here again to see it better prevIsReversed?

src/App.tsx
```diff
@@ -15,47 +15,119 @@ export const goodsFromServer = [
...
+            return (
+              <div>
+                <div
+                  data-cy="Good"
+                  key={good}
+                  className="box column is-info is-rounded mb-3"
+                >
+                  {good}
+                </div>
+              </div>
```
hajnaloltyan, line 128: it should be a list and not a div

hajnaloltyan (changes requested):
Good job, just a few cosmetic issue - in my opinion, could be wrong

src/App.tsx
VinnieJ-2k20, line 53: Follow React naming convention
Also, add generic type if you use custom or non-primitive data types.
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
VinnieJ-2k20: Yep, make it ul>li list

src/App.tsx
mayuces: I corrected thesee but apparentley forgot to push

VinnieJ-2k20 (approved):

PR #486, opened 2022-11-07
list of goods with useState hook was made

igor-1987:
[DEMO LINK](https://<your_account>.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,120 @@ export const goodsFromServer = [
...
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const goods = getReorderedGoods(
+    goodsFromServer,
+    { sortType, isReversed },
+  );
+
+  const sortByAlphabet = () => setSortType(SortType.ALPABET);
+  const sortByLength = () => setSortType(SortType.LENGTH);
+  const toggleReverse = () => setIsReversed(!isReversed);
```
Andr1yK, line 67: If you are based on the previous value, then you need to use functional set state

Andr1yK (changes requested):

src/App.tsx
igor-1987: > If you are based on the previous value, then you need to use functional set state

Initial values of state were located in useState initial values. We use "set state" through these methods (as setSortType, setIsReversed) and they look for initial value in its part of state to change it.
useState(initialState) keeps these values for us. Or may be, I didn't understand the comment.

igor-1987: > If you are based on the previous value, then you need to use functional set state

Done!

src/App.tsx
```diff
@@ -55,16 +55,16 @@ export function getReorderedGoods(
 
 export const App: React.FC = () => {
   const [isReversed, setIsReversed] = useState(false);
-  const [sortType, setSortType] = useState(SortType.NONE);
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
```
Andr1yK, line 58: 🔥

Andr1yK (approved):

PR #498, opened 2022-11-16
react. list of goods_hooks

boriskerekesha:


vlad-demchuk (changes requested):
add a demo link

boriskerekesha:
Hi, As I understandю, here I need to use useEffect ,to hide the button, but I don't understand how. https://boriskerekesha.github.io/react_list-of-goods-hooks/

PR #509, opened 2022-11-24
Hooked list of goods created

vadymbaranov:
[DEMO LINK](https://Largoxoxo.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,10 +63,6 @@ export const App: React.FC = () => {
   );
   const isOrderChanged = isReversed || sortType !== SortType.NONE;
 
-  useEffect(() => {
-
-  });
-
   return (
     <div className="section content">
       <div className="buttons">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const orderedGoods = getReorderedGoods(
+    goodsFromServer, { sortType, isReversed },
+  );
+  const isOrderChanged = isReversed || sortType !== SortType.NONE;
+
+  useEffect(() => {
+
+  });
```
Kry-Gin, line 68: Was it necessary to use useEffect in this code?

Kry-Gin (approved):
I have only question about useEffect. Otherwise you have done great job! (͠≖ ͜ʖ͠≖)👌

src/App.tsx
Largoxoxo: > Was it necessary to use useEffect in this code?

Oopsie, no) I simply forgot to delete it. Thanks for pointing out 👍

G-YaSha (approved):
Good job 👍

PR #516, opened 2022-11-26
solution

viktoriia-ovcharenko:
[DEMO LINK](https://viktoriia-ovcharenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,47 +16,119 @@ export const goodsFromServer = [
...
+        return good1.length - good2.length;
+
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.NONE:
+        return 0;
+
+      default:
+        throw new Error('Type is not valid');
```
referee1988, line 48:
```suggestion
      case SortType.NONE:
      default:
        return 0;
```
In this case, it's better not to throw an exception, but just return 0.
An exception will cause the application to break, and a return of 0 will work fine.
Giving the user the wrong sort is better than crashing the app with an error.

referee1988 (approved):
Great job!
Additionally, think about whether to throw an exception.

src/App.tsx
```diff
@@ -15,47 +16,119 @@ export const goodsFromServer = [
...
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((good1, good2) => {
```
mykola-koval, line 36:
```suggestion
  const visibleGoods = [...goods].sort((good1, good2) => {
```

mykola-koval (approved):

PR #519, opened 2022-11-27
hooks

Svitlana-Yudina:
[DEMO LINK](https://Svitlana-Yudina.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,134 @@ export const goodsFromServer = [
...
+
+  const reverse = () => {
+    setRevers(!isReversed);
+  };
+
+  const reset = () => {
+    setSortType(SortType.NONE);
+    setRevers(false);
+  };
+
```
Oleksandr-Stryzhavchuk, line 86: I think by writing all these functions separately, you are writing a lot of redundant code. it would be better to put them directly in the onclick

Oleksandr-Stryzhavchuk (approved):
in general, everything is cool, only one remark:)

src/App.tsx
Svitlana-Yudina: We have opposite opinions on this matter))) We will wait for the mentor's decision)

src/App.tsx
```diff
@@ -15,47 +16,134 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  ALPHABET,
+  LENGTH,
+  NONE,
+}
+
+// Use this function in the render to prepare goods
```
mykola-koval, line 25: let's remove comments

src/App.tsx
```diff
@@ -15,47 +16,134 @@ export const goodsFromServer = [
...
+export function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  visibleGoods.sort((good1, good2) => {
```
mykola-koval, line 35:
```suggestion
  const visibleGoods = [...goods].sort((good1, good2) => {
```

src/App.tsx
```diff
@@ -15,47 +16,134 @@ export const goodsFromServer = [
...
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
mykola-koval, line 54: 😮‍💨

src/App.tsx
mykola-koval: both your opinions are valid, there's no single best approach to writing code in programming.
For production code the functions (except reset) are probably too simple to move them from onClick, but we're learning here so that's a good approach too.

mykola-koval (changes requested):

mykola-koval (approved):

PR #549, opened 2022-12-02
Solution

viktor-kotik:
[DEMO LINK](https://viktor-kotik.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -16,47 +16,82 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const visibleGoods = [...goodsFromServer];
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortBy, setSortBy] = useState(0);
+
+  if (sortBy === 1) {
```
loralevitska, line 23: avoid magic numbers. Use enum and meaningful names for sorting

src/App.tsx
```diff
@@ -16,47 +16,82 @@ export const goodsFromServer = [
...
       <div className="buttons">
         <button
+          onClick={() => setSortBy(1)}
           type="button"
-          className="button is-info is-light"
+          className={
+            sortBy === 1
+              ? 'button is-info'
+              : 'button is-info is-light'
+          }
```
loralevitska, line 47: use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

src/App.tsx
```diff
@@ -16,47 +16,82 @@ export const goodsFromServer = [
...
-        >
-          Reset
-        </button>
+        {
+          (sortBy > 0 || isReversed === true)
+            ? (
+              <button
+                onClick={() => {
+                  setIsReversed(false); setSortBy(0);
+                }}
```
loralevitska, line 82: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
good start! Let's improve

src/App.tsx
```diff
@@ -15,48 +16,104 @@ export const goodsFromServer = [
...
+          className={classNames(
+            {
+              'button is-info is-light':
+                sortBy !== SortType.ALPHABET,
+            },
+            {
+              'button is-info':
+                sortBy === SortType.ALPHABET,
+            },
+          )}
```
loralevitska, line 64: If I understood you correctly, you want to always have some styles and add only one if necessary, then it is better to write it like this

```suggestion
          className={classNames('button', 'is-info', 
            {
              'is-light':
                sortBy !== SortType.ALPHABET,
            },
          )}
```

loralevitska (approved):

src/App.tsx
viktor-kotik: Yes thanks

PR #551, opened 2022-12-03
List of goods hooks

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +15,119 @@ export const goodsFromServer = [
...
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, sortBy] = useState(SortType.NONE);
+  const [isReversed, reverseArray] = useState(false);
+  const [isReset, resetVisible] = useState(false);
```
TarasHoliuk, line 65: no need to create the separate `isReset` state variable
you can calculate it from `isReversed` and `sortType`

TarasHoliuk (changes requested):
GJ! only one fix is needed

src/App.tsx
juliiaap: fixed

loralevitska (approved):
good job!

PR #560, opened 2022-12-16
hooks translated

metinbicaksiz:
[DEMO LINK](https://metinbicaksiz.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +15,86 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
```
jstmpelowycz, line 21:
```suggestion
  None,
  Alphabetically,
  ByLength,
```

src/App.tsx
```diff
@@ -15,47 +15,86 @@ export const goodsFromServer = [
...
+    visibleGoods.reverse();
+  }
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sortType !== SortType.ALPHABET ? 'is-light' : ''}`}
```
jstmpelowycz, line 51: just a suggestion: let's use [classnames](https://www.npmjs.com/package/classnames) here
using ternary and returning an empty string is not the best approach here

src/App.tsx
```diff
@@ -15,47 +15,86 @@ export const goodsFromServer = [
...
+          onClick={() => setSortType(SortType.ALPHABET)}
         >
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={`button is-success ${sortType !== SortType.LENGTH ? 'is-light' : ''}`}
```
jstmpelowycz, line 59: the same with classnames (if you want to)

src/App.tsx
```diff
@@ -15,47 +15,86 @@ export const goodsFromServer = [
...
+          onClick={() => setSortType(SortType.LENGTH)}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`button is-warning ${isReversed === false ? 'is-light' : ''}`}
```
jstmpelowycz, line 67: here as well, plus the following

src/App.tsx
```diff
@@ -15,47 +15,86 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {
+          (visibleGoods[0] !== goodsFromServer[0]) && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => {
+                setSortType(SortType.NONE);
+                setIsReversed(false);
```
jstmpelowycz, line 80: let's move this callback out in a dedicated function so that we don't write it right here (a kind of best practices)

src/App.tsx
jstmpelowycz:
```suggestion
          className={`button is-warning ${!isReversed ? 'is-light' : ''}`}
```

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,102 @@ export const goodsFromServer = [
...
+  };
+
+  const reverseIt = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const resetIt = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
```
jstmpelowycz, line 62: let's stick to naming convention
```
  const handleSortAlphabetically = () => {
    ...
  };

    ...
  };

  const handle... = () => {
    ...
  }

  ...
```

src/App.tsx
jstmpelowycz: uppercase values refers to some global variables like
```
const NUMBERS_IN_WEEK = 7;
```
for enum, it's better to use capitalized ones

jstmpelowycz (changes requested):

polosanya (approved):
Well done!

PR #568, opened 2022-12-25
Solution

Julia2063:
[DEMO LINK](https://Julia2063.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const copyGoods = [...goodsFromServer];
+  const [sortType, setsortType] = useState(SortType.NONE);
+  const [isReversed, setReversed] = useState(false);
+
+  const getReorderedGoods = (goods: string[]) => {
+    goods.sort((g1, g2) => {
```
Anastasiia-Svintsova, line 31: Sort is a mutable method. You don't need to rewrite it to a new constant as you do it below with `let visibleGoods`.  You can just sort a copy of goodsFromServer without an additional function

src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map(
+            good => (
+              <li data-cy="Good" key={good}>
+                {good}
+              </li>
+            ),
+          )}
```
Anastasiia-Svintsova, line 130:
```suggestion
          {visibleGoods.map(good => (
            <li data-cy="Good" key={good}>
              {good}
            </li>
          ))}
```

Anastasiia-Svintsova (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  let copyGoods = [...goodsFromServer];
+  const [sortType, setsortType] = useState(SortType.NONE);
+  const [isReversed, setReversed] = useState(false);
```
polosanya, line 28:
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```
SetState function should be named according to the noun name. 
Also, don't forget about camelCase

polosanya, line 28:
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```
SetState function should be named according to the noun name. 
Also. don't forget about camelCase

polosanya, line 28:
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```
SetState function should be named according to the noun name. 
Also. don't forget about camelCase

src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+
+  copyGoods.sort((g1, g2) => {
```
polosanya, line 30: You can use useEffect here with sortType in the dependency array to sort the goods only when sortType changes.

src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+        return g1.length - g2.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    copyGoods.reverse();
+  }
```
polosanya, line 45: Same here, but with isReversed

src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+
+  const sortAlphabetically = () => {
+    setsortType(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    setsortType(SortType.LENGTH);
+  };
+
+  const reverce = () => {
```
polosanya, line 55:
```suggestion
  const reverse = () => {
```
Btw, it's called the same as array's method, so it would be better to choose a bit another name

polosanya, line 55:
```suggestion
  const reverse = () => {
```
Btw, it's called the same as array's method, so it would be better to choose a bit another name

polosanya, line 55:
```suggestion
  const reverse = () => {
```
Btw, it's called the same as array's method, so it would be better to choose a bit another name

src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+    setReversed(!isReversed);
+  };
+
+  const reset = () => {
+    setsortType(SortType.NONE);
+    setReversed(false);
+    copyGoods = goodsFromServer;
```
polosanya, line 62:
```suggestion
    copyGoods = [...goodsFromServer];
```

polosanya (changes requested):
Almost perfect!

src/App.tsx
loralevitska: not fixed

loralevitska (changes requested):
There is still a little bit left, pay attention to the uncorrected comments (I left them unresolved). This will help avoid unnecessary calculations in your code. If you have any questions, please ask them in fe_chat

src/App.tsx
Julia2063: Same here

src/App.tsx
Julia2063: sorry, we had a conversation with previous checker, and solved to stay this points like its were.

loralevitska (approved):

PR #570, opened 2022-12-26
Solution 1

JohnRust12:
[DEMO LINK](https://JohnRust12.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,49 +15,120 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+// Use this function in the render to prepare goods
```
polosanya, line 29: Do you really need the comments? Your code should be understandable without any comments

src/App.tsx
```diff
@@ -14,49 +15,120 @@ export const goodsFromServer = [
...
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
polosanya, line 39: Same with console.log. Remove it

src/App.tsx
```diff
@@ -14,49 +15,120 @@ export const goodsFromServer = [
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {shouldShowResetButton && (
+          <button
+            type="button"
+            className="button is-danger is-rounded is-light"
+            onClick={() => {
```
polosanya, line 116: Create a separate function to handle this click.
Also, would be great to handle all the clicks with separate functions

polosanya (changes requested):
Let's polish it up)

src/App.tsx
JohnRust12: separated

src/App.tsx
JohnRust12: deleted

loralevitska (approved):
good job!

PR #572, opened 2022-12-30
Solution react_list-of-goods-hooks

anton-martyniuk1:
[DEMO LINK](https://anton-martyniuk1.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -16,46 +16,88 @@ export const goodsFromServer = [
...
+      goods.sort((g1, g2) => g1.length - g2.length);
+      break;
+
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    goods.reverse();
+  }
```
SanyaBratashchuk, line 55: create separate function for sort goods. 
So that it works like that `const visibleGoods = sortGoods(goods)`
[Optional] use `useMemo` here

src/App.tsx
```diff
@@ -16,46 +16,88 @@ export const goodsFromServer = [
...
+          className={`button is-info ${sortType !== 'alphabet' && 'is-light'}`}
+          onClick={sortByAlphabet}
         >
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={`button is-success ${sortType !== 'length' && 'is-light'}`}
```
SanyaBratashchuk, line 70: use classname or clsx libraries

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #587, opened 2023-01-05
solution

Fynkzia:
[DEMO LINK](https://Fynkzia.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -14,49 +15,127 @@ export const goodsFromServer = [
...
+  );
+  const resetVisibility = (sortType !== SortType.NONE) || isReversed;
+
+  const handleSortAlphabet = () => {
+    setSortType(SortType.ALPHABET);
+  };
+
+  const handleSortLength = () => {
+    setSortType(SortType.LENGTH);
+  };
```
superpooperxxx, line 77: You can combine those functions into one, as they are both setting a type, so you can pass the type as parameter.
```
  const handleSortValue = (type) => {
     setSortType(type);
  };
```

superpooperxxx (approved):
🔥

src/App.tsx
```diff
@@ -14,49 +15,127 @@ export const goodsFromServer = [
...
+}
 
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+  const visibleGoods = getReorderedGoods(
+    goodsFromServer,
+    { isReversed, sortType },
+  );
+  const resetVisibility = (sortType !== SortType.NONE) || isReversed;
```
Ihor-Karpyn, line 69: This name looks like of function, if it's boolean variable start name from is `isResetButtonVisible`

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #619, opened 2023-01-08
Solution

TetKopot:
[DEMO LINK](https://TetKopot.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


anastasiia-bilska (approved):
well done!

src/App.tsx
```diff
@@ -1,6 +1,72 @@
...
+      >
+        {theme.palette.mode === 'dark'
+          ? <Brightness7Icon /> : <Brightness4Icon />}
+      </IconButton>
+    </Box>
+  );
+}
+
+export default function ToggleColorMode() {
+  const [mode, setMode] = React.useState<'light' | 'dark'>('light');
```
valerii-derkach, line 45: create type or enum for mode types

src/App.tsx
```diff
@@ -15,48 +81,122 @@ export const goodsFromServer = [
...
+  goodsOnPage.sort((first, second) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return first.localeCompare(second);
+
+      case SortType.LENGTH:
+        return first.length - second.length;
+
+      default:
+        return 0;
```
valerii-derkach, line 106:
```suggestion

     case SortType.NONE:
      default:
        return 0;
```

best practice is to handle all cases

src/App.tsx
```diff
@@ -15,48 +81,122 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    goodsOnPage.reverse();
+  }
+
+  return goodsOnPage;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsREversed] = useState(false);
```
valerii-derkach, line 118:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +81,122 @@ export const goodsFromServer = [
...
+  const goodsOnPage = [...goods];
+
+  goodsOnPage.sort((first, second) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return first.localeCompare(second);
+
+      case SortType.LENGTH:
+        return first.length - second.length;
+
```
valerii-derkach, line 104: better to sort inside each case to avoid redundant actions when it is no need to do anything

valerii-derkach (changes requested):
Good job!

few comments to fix:

look at your theme mode container on the page - does it work as you expected? looks like it should work on the whole page, but it doesn't

and what about a separate file for each component?

src/App.tsx
```diff
@@ -15,48 +17,129 @@ export const goodsFromServer = [
...
+      goodsOnPage.sort((first, second) => {
+        return first.length - second.length;
+      });
+      break;
+
+    case SortType.NONE:
+      break;
+
+    default:
+      break;
```
ivan-kadykalo, line 50:
```suggestion
    case SortType.NONE:
    default:
      break;
```

ivan-kadykalo (approved):
Good job 🔥

PR #624, opened 2023-01-09
React List of Goods - Hooks

YanaEvaGonchar:
- [DEMO LINK](https://YanaEvaGonchar.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
+import cn from 'classnames';
+
 import './App.scss';
 
 export const goodsFromServer = [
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((good1, good2) => good1.length - good2.length);
+      break;
+
+    case SortType.NONE:
+    default:
+      break;
+  }
```
romanblyashuk94, line 49:
```suggestion
  switch (sortType) {
    case SortType.ALPHABET:
      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
      break;

    case SortType.LENGTH:
      visibleGoods.sort((good1, good2) => good1.length - good2.length);
      break;

    default:
  }
```

romanblyashuk94 (approved):
good job 👍

src/App.tsx
valerii-derkach: it is better to handle all cases from enum to avoid unexpected behavior in the future, - what if another developer passes this unusable value to your function, and forgets to check if this case is handled? So best practice is to handle this case also and just break, and do not leave space for mistakes

src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
-          className="button is-warning is-light"
+          className={cn(
+            'button',
+            'is-warning',
+            { 'is-light': !isReversed },
+          )}
+          onClick={() => {
+            setIsReversed(() => {
+              return !isReversed;
+            });
```
valerii-derkach, line 103: move logic to method

src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
+        {
+          (sortType !== SortType.NONE || isReversed) && (
+            <button
+              type="button"
+              className="button is-danger"
+              onClick={() => {
+                setSortType(SortType.NONE);
+                setIsReversed(false);
+              }}
+            >
```
valerii-derkach, line 118: create a variable for the condition and move onClick logic to method

src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
           type="button"
-          className="button is-success is-light"
+          className={cn(
+            'button',
+            'is-success',
+            { 'is-light': sortType !== SortType.LENGTH },
+          )}
+          onClick={() => {
+            setSortType(SortType.LENGTH);
+          }}
```
valerii-derkach, line 88: move logic to method

src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {
+          getReorderedGoods(goodsFromServer, { sortType, isReversed })
+            .map((good) => {
```
valerii-derkach, line 128: create a variable

src/App.tsx
```diff
@@ -15,48 +17,122 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn(
+            'button',
+            'is-info',
```
valerii-derkach, line 69:
```suggestion
            'button is-info',
```

it is ok to write these two classes in one string

valerii-derkach (changes requested):

src/App.tsx
YanaEvaGonchar: done

src/App.tsx
YanaEvaGonchar: done

src/App.tsx
YanaEvaGonchar: done

src/App.tsx
YanaEvaGonchar: done

src/App.tsx
YanaEvaGonchar: done

valerii-derkach (approved):

PR #632, opened 2023-01-10
Develop

SalehAkaJim:
[DEMO LINK](https://SalehBUD.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
+
+  visibleGoods.sort((good, nextGood) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good.localeCompare(nextGood);
+
+      case SortType.LENGTH:
+        return good.length - nextGood.length;
+
+      case SortType.NONE:
```
lefteris-L, line 44: It's better to check this with an if statement so you start a sort loop only if you have to

src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
           Reverse
         </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed !== false) && (
```
lefteris-L, line 113:
```suggestion
        {(sortType !== SortType.NONE || isReversed) && (
```

lefteris-L (approved):
Very nice work 🔥  just a couple of suggestions

src/App.tsx
naziamoff: @lefteris-L I'd say that switch works better here. Limited number of conditions and we know them all exactly

src/App.tsx
naziamoff: I believe that we should separate this variable. If you need to change the condition later - you'll need to look for this exact piece of code. Much more easy if it is a variable

naziamoff (approved):

PR #638, opened 2023-01-11
react_list-of-goods-hooks

JzSol:
[DEMO LINK](https://JzSol.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,49 +16,107 @@ export const goodsFromServer = [
...
+  isReversed: boolean;
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((prev, curr) => {
```
Grayfox147, line 36: why not just switch here? it would be more clear, an perfect for the situation.

src/App.tsx
```diff
@@ -15,49 +16,107 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+}
+
+export const App: React.FC<{}> = () => {
+  const [isReversed, reverseList] = useState(false);
+  const [sortType, sortItems] = useState(SortType.NONE);
+
+  const goods = getReorderedGoods(goodsFromServer, { isReversed, sortType });
+  const activeRemoveButton = isReversed || sortType !== SortType.NONE;
```
Grayfox147, line 62: great job saving the checking conditions in a variable!

Grayfox147 (commented):
Just minor comments but really good job overall!

src/App.tsx
naziamoff: switch is better here, agree with @Grayfox147

src/App.tsx
naziamoff: 🔥 🔥 🔥 
`isRemoveButtonActive` would be better since it is a boolean value

naziamoff (approved):

PR #639, opened 2023-01-11
solution

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,121 @@ export const goodsFromServer = [
...
+  isReversed: boolean,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
+
+  if (sortType === SortType.ALPHABET) {
```
Grayfox147, line 38:
```suggestion
  if (sortType) {
    visibleGoods.sort((g1, g2) => {
      switch (sortType) {
        case SortType.ALPHABET:
          return g1.localeCompare(g2);

        case SortType.LENGTH:
          return g1.length - g2.length;

        default:
          return 0;
      }
    });
  }
```

src/App.tsx
```diff
@@ -15,47 +16,121 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
Grayfox147, line 118: (isReversed || sortType !== SortType.NONE 
So this long comparison checking would look better inside a variable, and then you can use the variable instead!

Grayfox147 (commented):
Excellent job! just a couple of suggestions but overall is an amazing job! really smart implementation!

src/App.tsx
mmahfuzi: Done :)

mmahfuzi:
> Excellent job! just a couple of suggestions but overall is an amazing job! really smart implementation!

Thanks :)

src/App.tsx
mmahfuzi: Done

src/App.tsx
```diff
@@ -15,47 +16,126 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+// Use this function in the render method to prepare goods
```
naziamoff, line 25: you should remove all the unnecessary comments

src/App.tsx
```diff
@@ -15,47 +16,126 @@ export const goodsFromServer = [
...
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
+
+  if (sortType) {
+    visibleGoods.sort((g1, g2) => {
```
naziamoff, line 39: not really good idea to name it like `g1` or `g2` or whatever. Names should be clear, even in the callbacks

src/App.tsx
```diff
@@ -15,47 +16,126 @@ export const goodsFromServer = [
...
+    setSort(SortType.LENGTH);
+  };
+
+  const reset = () => {
+    setIsReversed(false);
+    setSort(SortType.NONE);
+  };
+
+  const visibleGoods = getReorderedGoods(goodsFromServer, sortType, isReversed);
+  const resetCheck = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 82: it is a boolean variable, so this should have a boolean name

naziamoff (approved):
good job. But pay attention to the naming

src/App.tsx
mmahfuzi: A name just came to my mind which should be better, I will use that.

PR #643, opened 2023-01-11
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,121 @@ export const goodsFromServer = [
...
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const visibleGoods = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
+  const toggleSortType = (type: SortType) => {
+    return sortType === type
+      ? setSortType(SortType.NONE)
+      : setSortType(type);
+  };
```
nurdanokcu, line 57: I liked the toggle approach but since handleSortByAlphab and handleSortByLength functions are already created, toggle now seems like a redundant function.

nurdanokcu (approved):
I added my suggestion , 
I am approving and you can decide whether my suggestion makes sense.

src/App.tsx
```diff
@@ -15,48 +16,121 @@ export const goodsFromServer = [
...
+
+  const handleSortByLength = () => (
+    toggleSortType(SortType.LENGTH)
+  );
+
+  const handleReverse = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const handleReset = () => {
```
naziamoff, line 71:
```suggestion
  const sortAlphabetically = () => (
    toggleSortType(SortType.ALPHABET)
  );

  const sortByLength = () => (
    toggleSortType(SortType.LENGTH)
  );

  const reverse = () => {
    setIsReversed(!isReversed);
  };

  const reset = () => {
```

src/App.tsx
naziamoff: agree with @nurdanokcu on this one

naziamoff (approved):
well done! Just some minor suggestions

PR #644, opened 2023-01-11
solution

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,10 +27,8 @@ export function getReorderedGoods(
   sortType: SortType,
   isReversed: boolean,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
   if (sortType) {
     visibleGoods.sort((g1, g2) => {
       switch (sortType) {
@@ -49,9 +47,6 @@ export function getReorderedGoods(
   if (isReversed) {
     visibleGoods.reverse();
   }
-
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
 
   return visibleGoods;
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
           Reverse
         </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== NONE) && (
```
kenvastolord, line 105: I would recommend save (isReversed || sortType !== NONE) into a variable and then use the variable here, just that, everything from that point looks wonderful.

kenvastolord (commented):
Excellent job on that project.

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {
+          visibleGoods.map(good => (
+            <li data-cy="Good" key={good}>
```
naziamoff, line 121:
```suggestion
        {visibleGoods.map(good => (
            <li data-cy="Good" key={good}>
```

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReverseState] = useState(false);
```
naziamoff, line 56: no need to think of the names for states. It always works like `[value, setValue]`
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

naziamoff (approved):
well done, but I recommend you to fix a couple of things including the @kenvastolord comment

PR #646, opened 2023-01-12
final

KareemAchkar:
(https://KareemAchkar.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,153 @@ export const goodsFromServer = [
...
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  // To avoid the original array mutation
```
kenvastolord, line 34: It is not important, but just make sure you remove the comments before git push the code.
It good practice comment your code, well done!,  but just to mention, in production, the code should not have comments. I know this is not production, but just leave it without them.

src/App.tsx
```diff
@@ -15,47 +16,153 @@ export const goodsFromServer = [
...
+  //   this.setState({
+  //     sortType: SortType.LENGTH,
+  //   });
+  // };
+
+  // handleReverse = () => {
+  //   this.setState((state) => ({
+  //     isReversed: !state.isReversed,
+  //   }));
+  // };
```
kenvastolord, line 88: the same here

src/App.tsx
```diff
@@ -15,47 +16,153 @@ export const goodsFromServer = [
...
+
+  const resetHandle = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
+
+  // render(): React.ReactNode {
+  //   const { isReversed, sortType } = this.state;
```
kenvastolord, line 96: the same with this comment.

src/App.tsx
```diff
@@ -15,47 +16,153 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
kenvastolord, line 143: I would recommend save this (isReversed || sortType !== SortType.NONE) into a variable befor the return() and use the variable here.

src/App.tsx
naziamoff: @kenvastolord it is really important, since those comments in the real project can really confuse

src/App.tsx
naziamoff: agree with @kenvastolord

meforoma (approved):
well done 👍🏼

PR #647, opened 2023-01-12
sol1

dolnaro:
[DEMO LINK](https://dolnaro.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,118 @@ export const goodsFromServer = [
...
           Reverse
         </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType || isReversed) && (
```
Rijo-J, line 112: Could be stored in a variable i.e. `isNotDefault`

Rijo-J (approved):
Very nice! Legit!! 👍 🚀

src/App.tsx
```diff
@@ -15,48 +16,118 @@ export const goodsFromServer = [
...
+    goodsFromServer, { isReversed, sortType },
+  );
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button', 'is-info', {
```
naziamoff, line 85: not necessary to make each class a separate string
```suggestion
          className={classNames('button is-info', {
```

src/App.tsx
naziamoff: agree with @Rijo-J, it is important to separate the variables from the code. Will be much easier to maintain later

naziamoff (approved):
great job!

src/App.tsx
dolnaro: I Will change it :)

PR #696, opened 2023-02-02
task was solved

ihor-soloviov:
[DEMO LINK](https://ihor-soloviov.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
+
+      case SortType.LENGTH:
+        return first.length - next.length;
+
+      case SortType.NONE:
+        return 0;
+
+      default:
+        throw new Error('Wrong sort type');
+        break;
```
Coyoote, line 45: No need to add break here, as the statements after `throw` won't be executed.

src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
+  const [reversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const sortAlphabetically = () => {
+    setSortType(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
```
Coyoote, line 66: You don't really need to make a separate function for both set functions as you can call them in onClick handler

Coyoote (approved):
Good job! Left a few comments on improving your code.

src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
+
+  if (reversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [reversed, setReversed] = useState(false);
```
ArturErmolenko, line 57:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +16,125 @@ export const goodsFromServer = [
...
+  const reverseList = () => {
+    setReversed(current => !current);
+  };
+
+  const resetList = () => {
+    setSortType(SortType.NONE);
+    setReversed(false);
+  };
+
+  const isResetBtnClicked = Boolean(sortType !== SortType.NONE || reversed);
```
ArturErmolenko, line 77:
```suggestion
  const shouldShowResetButton = Boolean(sortType !== SortType.NONE || reversed);
```

ArturErmolenko (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
+    setIsReversed(current => !current);
+  };
+
+  const resetList = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
+
+  const shouldShowResetButton = Boolean(sortType !== SortType.NONE
+    || isReversed);
```
alexandra-protyanova, line 69:
```suggestion
  const isResetButtonVisible = sortType !== SortType.NONE
    || isReversed;
```

alexandra-protyanova: Both operands return `boolean`, so there is no need to convert it

src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        { shouldShowResetButton && (
```
alexandra-protyanova, line 110:
```suggestion
        {shouldShowResetButton && (
```

alexandra-protyanova (approved):
As for me it's better to create component for `GoodsList` to make `App.tsx` structure less complicated 😊

PR #702, opened 2023-02-02
goods hooks was done

OlyaZaiets:
[DEMO LINK](https://OlyaZaiets.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames(
+            'button is-info',
+            { 'is-light': sortType !== SortType.ALPHABET },
+          )}
+          onClick={() => setSortType(SortType.ALPHABET)}
```
vladyslav78292, line 78: I think it would be better to write a function that set sort type by length and alphabet, and just call it here

vladyslav78292 (approved):
good job!

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
+        throw new Error('Unexpected sort type!');
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
alexandra-protyanova, line 55: I don't think you need to console it

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const showResetButton = sortType !== SortType.NONE || isReversed;
```
alexandra-protyanova, line 63:
```suggestion
  const isResetButtonVisible = sortType !== SortType.NONE || isReversed;
```

src/App.tsx
alexandra-protyanova: agree, let us move this logic to separate function

alexandra-protyanova (changes requested):
It is better to create more components, to make your components more readable ☺️

PR #719, opened 2023-02-05
solution

miamarichka:
[DEMO LINK](https://miamarichka.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
+import classNames from 'classnames';
 import './App.scss';
+import { GoodsList } from './GoodList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,49 +17,108 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReversed] = useState(false);
+
+  const reverse = () => {
+    setReversed(!isReversed);
```
agniya-i, line 56:
```suggestion
    setReversed((prevState) => !prevState);
```

src/App.tsx
```diff
@@ -15,49 +17,108 @@ export const goodsFromServer = [
...
+
+  const reset = () => {
+    setReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const goodsToRender
+      = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
+  const wasItSorted = sortType === SortType.NONE && isReversed === false;
```
agniya-i, line 75:
```suggestion
  const isSorted = sortType !== SortType.NONE && isReversed;
```
it is better to return `true` from the expression if you have a such boolean check, it can be confusing

agniya-i (changes requested):

src/App.tsx
miamarichka: const isSorted = sortType !== SortType.NONE && isReversed;
this suggestion isnt work in this way, it shoul be opposite, by the way I done it correct)

src/App.tsx
```diff
@@ -15,49 +17,108 @@ export const goodsFromServer = [
...
+
+  const reset = () => {
+    setReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const goodsToRender
+      = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
+  const wasItSorted = sortType === SortType.NONE && !isReversed;
```
polosanya, line 75:
```suggestion
  const isSorted = sortType !== SortType.NONE || isReversed;
```

src/App.tsx
```diff
@@ -15,49 +17,108 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {!wasItSorted && (
```
polosanya, line 110:
```suggestion
        {isSorted && (
```
It's more logical to check for some true value, not its denial

polosanya (approved):

PR #723, opened 2023-02-08
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
-import 'bulma/css/bulma.css';
-import './App.scss';
+import classNames from 'classnames';
+import React, { useState } from 'react';
 
-export const goodsFromServer = [
+import { GoodsList } from './components/GoodsList';
+import { ListControl } from './components/ListControl';
+
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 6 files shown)


slavanosatskyi (changes requested):
Your tests are falling. Seems that you forgot to add data-cy="Good" data attribute to your goods. Let's try to fix it

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,13 @@
+export const GoodsList: React.FC<Props> = ({ goods }) => {
+  return (
+    <ul className="goods-list">
+      {goods.map(good => (
+        <li className="goods-list__item">{good}</li>
```
SanyaBratashchuk, line 5: add keys

src/App.tsx
```diff
@@ -15,49 +17,64 @@ export const goodsFromServer = [
...
+          return a.length - b.length;
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
SanyaBratashchuk, line 47: create separate function for sorting todos

SanyaBratashchuk (changes requested):

yaroslav-furhala:
Hi! I reinstall node modules, but it`s doesn`t help fix this problem, I guess it`s cypress bug, like we said. I add data-cy="Good" attribute but it doesn`t help too.
image

src/App.tsx
```diff
@@ -15,49 +17,64 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
-      </ul>
+    <div className={classNames('App', {
+      'App--before': !isListVisible,
+      'App--after': isListVisible,
+    })}
+    >
+      {!isListVisible
```
Volodymyr-Melnyk-Fron10der, line 55: It's better to handle truth case first to reach more clear and readable code

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,19 @@
...
+          {good}
+        </li>
+      ))}
+    </ul>
+  );
+};
+
+type Props = {
+  goods: string[]
+};
```
Volodymyr-Melnyk-Fron10der, line 19: Try to keep well-known component structure

```suggestion
type Props = {
  goods: string[]
};

const GoodsList: React.FC<Props> = ({ goods }) => {
  return (
    <ul className="goods-list">
      {goods.map(good => (
        <li
          key={good}
          data-cy="Good"
          className="goods-list__item"
        >
          {good}
        </li>
      ))}
    </ul>
  );
};

export default GoodsList
```

src/App.tsx
```diff
@@ -15,49 +17,64 @@ export const goodsFromServer = [
...
+  const [isListVisible, setisListVisible] = useState(false);
+  const [isReversed, setisReversed] = useState(false);
+  const [sortBy, setSortBy] = useState('');
+
+  const reset = () => {
+    setisReversed(false);
+    setSortBy('');
+  };
+
+  const visibleGoods = [...goodsFromServer];
```
Volodymyr-Melnyk-Fron10der, line 30: I think Oleksandr did mean to make smth like that:

And your function should be placed for example here `src -> helpers -> get-sorted-goods.ts`
```suggestion
  const visibleGoods = getSortedGoods(goodsFromServer, sortBy, isReversed);
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good, but check my comments bellow please

yaroslav-furhala:
1. (It's better to handle truth case first to reach more clear and readable code) - Fixed!
2. (Try to keep well-known component structure) - Done!
3. (I think Oleksandr did mean to make smth like that:) - I have to create separet class component ?

Volodymyr-Melnyk-Fron10der:
Why class component? Just a simple function getSortedGoods, which will contain all this code 
```
  if (sortBy) {
    visibleGoods.sort((a, b) => {
      switch (sortBy) {
        case 'alpha':
          return a.localeCompare(b);
        case 'length':
          return a.length - b.length;
        default:
          return 0;
      }
    });
  }

  if (isReversed) {
    visibleGoods.reverse();
  }
```
And please if you have any questions ask them in the Slack, not here

Volodymyr-Melnyk-Fron10der (changes requested):
Still need a fix

PR #732, opened 2023-02-12
Solution

offsset:
https://offsset.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,49 +16,115 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, sortTypeF] = useState(SortType.NONE);
```
OleziO, line 51: State setter names should be prefixed with set like this
`const [sortType, setSortType] = useState(SortType.NONE);`
Check other similar cases

src/App.tsx
```diff
@@ -15,49 +16,115 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={
+            classNames(
+              'button is-info',
+              { 'is-light': isReversed === false },
```
OleziO, line 96: Use `!isReversed`, its the same as `isReversed === false`

OleziO (approved):

PR #741, opened 2023-02-25
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
+    setIsReversed(false);
+  };
+
+  const handleLengthSort = () => {
+    setSortType(SortType.LENGTH);
+    setIsReversed(false);
+  };
+
+  const handleReverse = () => {
+    setIsReversed(!isReversed);
```
ivbutsykin, line 72: It's best practice to rely on actual state when use `setState` with relay on value
```suggestion
    setIsReversed(prev => !prev);
```

src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
+      case SortType.ALPHABET:
+        return item1.localeCompare(item2);
+
+      case SortType.LENGTH:
+        return item1.length - item2.length;
+
+      default:
+        return 0;
+    }
+  });
```
ivbutsykin, line 47: I'd prefer to move switch case before `visibleGoods.sort` to iterate like this:
```suggestion
switch (sortType) {
      case SortType.ALPHABET:
        return visibleGood.sort(...);
      case SortType.LENGTH:
        return visibleGoods.sort(...);
      default:
        return visibleGoods;
  });
```
Because your sort type in function doesn’t change, but switch case continue check on each iteration.

src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+  const goods = getReorderedGoods(goodsFromServer, { sortType, isReversed });
```
ivbutsykin, line 59: Here should be `useMemo` to improve performance a little bit.

You also can move getReorderedGoods inside component and wrap in `useCallback` but it depends on taste

ivbutsykin (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,131 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(false);
+
+  const getReorderedGoods = useCallback(
+    (
+      goods: string[],
+      {
+        sortType: reorderSortType,
+        isReversed: reorderIsReversed,
+      }: ReorderOptions,
+    ) => {
```
anstsot, line 41: this part is hard to read, you can improve it a little bit
```suggestion
  const getReorderedGoods = useCallback((
      goods: string[],
      {
        sortType: reorderSortType,
        isReversed: reorderIsReversed,
      }: ReorderOptions,
    ) => {
```

anstsot (approved):
Great job, approved!

PR #744, opened 2023-02-27
solution

TetianaKoval:
[DEMO LINK](https://TetianaKoval.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +16,126 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+
+  // eslint-disable-next-line no-console
+  // console.log(sortType, isReversed);
+}
+
+export function App() {
+  const [sortType, sort] = useState(SortType.NONE);
+  const [isReversed, reverse] = useState(false);
```
anstsot, line 60: it's better to rename the function that updates the state and use _set_. it's standard name convention and is better for readability
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```

anstsot (approved):
Great job, approved! You can add small improvement if you want

anstsot (approved):
Approved

PR #747, opened 2023-03-01
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
+    visibleGoods.sort((g1, g2) => {
+      return g1.localeCompare(g2);
+    });
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((g1, g2) => {
+      return g1.length - g2.length;
+    });
+  }
```
vlad-demchuk, line 46: the same, if we compare one value, it will be clearer to use switch

vlad-demchuk (approved):

PR #755, opened 2023-03-06
solution

webdevw:
[DEMO LINK](https://webdevw.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,128 @@ export const goodsFromServer = [
...
+                className="button is-danger is-light"
+                onClick={() => {
+                  setSortType(SortType.NONE);
+                  setIsReversed(false);
+                }}
+              >
+                Reset
+              </button>
+            )
+        }
```
anstsot, line 129: the piece can be changed for better readability
```suggestion
        {(sortType !== 'none' || isReversed) && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={() => {
              setSortType(SortType.NONE);
              setIsReversed(false);
            }}
          >
            Reset
          </button>
        )}
```

src/App.tsx
```diff
@@ -15,48 +16,128 @@ export const goodsFromServer = [
...
+    switch (sortType) {
+      case 'none':
+        return 0;
+      case 'alphabet':
+        return a.localeCompare(b);
+      case 'length':
+        return a.length - b.length;
+      default:
+        return 0;
+    }
```
anstsot, line 46: default and none cases can be merged. also let's use the power of typescript and check types better
```suggestion
    switch (sortType) {
        return 0;
      case SortTyp.ALPHABET:
        return a.localeCompare(b);
      case SortTyp.LENGTH:
        return a.length - b.length;
      case SortType.NONE:
      default:
        return 0;
    }
```

anstsot (approved):
Good job, approving!
You can add some improvements (check comments)

PR #763, opened 2023-03-08
result

novak-k:
[DEMO LINK](https://novak-k.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,130 @@ export const goodsFromServer = [
...
+      isReversed: prev.isReversed,
+    }));
+  };
+
+  const sortLength = () => {
+    secondState((prev) => ({
+      sortType: SortType.LENGTH,
+      isReversed: prev.isReversed,
+    }));
+  };
```
Viktor-Kost, line 77: you can join these function, because they do the same. Just pass a sortType as an argument

src/App.tsx
```diff
@@ -15,47 +16,130 @@ export const goodsFromServer = [
...
+  return visibleGoods;
+}
+
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export const App: React.FC<{}> = () => {
+  const [firstState, secondState] = useState<State>({
```
Viktor-Kost, line 60: you should change names a little bit according to naming useState variables 

Examples: 
```
const [sortType, setSortType] = useState('')
const [isValid, setIsValid] = useState(true)
```

src/App.tsx
```diff
@@ -15,47 +16,130 @@ export const goodsFromServer = [
...
+          onClick={sortLength}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={classNames('button is-warning',
+            { 'is-light': firstState.isReversed === false })}
```
Viktor-Kost, line 119:
```suggestion
            { 'is-light': !firstState.isReversed })}
```

src/App.tsx
```diff
@@ -15,47 +16,130 @@ export const goodsFromServer = [
...
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(firstState.sortType !== SortType.NONE
+          || firstState.isReversed === true)
```
Viktor-Kost, line 126:
```suggestion
        {(firstState.sortType !== SortType.NONE
          || firstState.isReversed)
```

Viktor-Kost (changes requested):
Nice!

Some moments need to be improved

src/App.tsx
novak-k: sure!

src/App.tsx
novak-k: yeah, naming is the hardest. )))

src/App.tsx
novak-k: done

src/App.tsx
```diff
@@ -15,47 +16,124 @@ export const goodsFromServer = [
...
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+      case SortType.LENGTH:
+        return a.length - b.length;
+      default:
+        return SortType.NONE;
+    }
+  });
```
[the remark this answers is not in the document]
novak-k: if I add sort in every case, I also need to add reverse there - which will increase the amount of code and confuse logic - otherwise it doesn't work properly 

`switch (sortType) {
    case SortType.ALPHABET:
      return visibleGoods.sort((a, b) => a.localeCompare(b)) ;
    case SortType.LENGTH:
    default:
      return true;
  }

  if (isReversed) {
    visibleGoods.reverse();
  }

  return visibleGoods;`

PR #784, opened 2023-03-19
:white_check_mark: implement listOfGoods using hooks

Bogdan-Mykhailov:
[DEMO LINK](https://Bogdan-Mykhailov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/components/Goods/Goods.tsx
+++ b/src/components/Goods/Goods.tsx
@@ -1,0 +1,18 @@
+import { FC } from 'react';
+import { createId } from '../../utils/uuid-creator';
+
+interface GoodsProps {
+  goods: string[];
+}
+
+export const Goods: FC<GoodsProps> = ({ goods }) => {
+  return (
+    <ul>
+      {goods.map(good => (
+        <li key={createId()} data-cy="Good">
+          {good}
+        </li>
+      ))}
+    </ul>
+  );
+};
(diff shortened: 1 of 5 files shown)


src/components/Goods/Goods.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { createId } from '../../utils/uuid-creator';
+
+interface GoodsProps {
+  goods: string[];
+}
+
+export const Goods: FC<GoodsProps> = ({ goods }) => (
+  <ul>
+    {goods.map(good => (
+      <li key={createId()} data-cy="Good">
```
OleksiiBlezniuk, line 11: I guess this function is a bit overkill, good will be enough here, but let it be as it is

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
+import { ButtonHTMLAttributes, FC } from 'react';
+import classNames from 'classnames';
+
+type Attributes = Omit<
+ButtonHTMLAttributes<HTMLButtonElement>,
+'className' | 'type'
+>;
```
OleksiiBlezniuk, line 7:
```suggestion
type Attributes = Omit<
  ButtonHTMLAttributes<HTMLButtonElement>,
  'className' | 'type'
>;
```

OleksiiBlezniuk: BTW, great idea to pass default button props to component

OleksiiBlezniuk (approved):

PR #790, opened 2023-03-21
List of goods hooks

Igor-Yaremchuk:
[List of goods hooks](https://Igor-Yaremchuk.github.io/react_list-of-goods-hooks/)


witflash (approved):
Great! I like it!

PR #805, opened 2023-03-22
react_list_of_goods-hooks solution

stanislavkulakovskyi:
[DEMO LINK](https://stanislavkulakovskyi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
+import { GoodsList } from './components/GoodsList';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+      break;
+
+    default:
+      break;
+  }
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
```
Yekatery, line 59: Here you could write shorter:
  if (isReversed) {
    visibleGoods.reverse();
  }

Yekatery (approved):
Well done!)

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
```
sergii-nosachenko, line 25: Always add values to enum

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
```
sergii-nosachenko, line 36: Add return type

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+      break;
+  }
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReverse] = useState(false);
```
sergii-nosachenko, line 62:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+type Props = {
+  goods: string[];
+};
+
+export const GoodsList: React.FC<Props> = ({ goods }) => {
+  return (
```
sergii-nosachenko, line 8: Useless `return`

sergii-nosachenko (changes requested):

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+  const sortByType = (type: SortType) => {
+    setSortType(type);
+  };
+
+  const reset = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
+
+  const showResetButton = sortType !== SortType.NONE || isReversed;
```
Andr1yK, line 78:
```suggestion
  const isResetButtonVisible = sortType !== SortType.NONE || isReversed;
```

Andr1yK, line 78:
```suggestion
  const shouldShowResetButton = sortType !== SortType.NONE || isReversed;
```

Andr1yK, line 78: Naming conventions for boolean variables and functions are important because they help make code more readable and easier to understand.

Boolean variables, which represent true or false values, are often used to store the state of a condition in a program. Starting their names with "is", "should", or "does" helps to indicate that they are boolean variables and makes it clear that their value represents a true or false condition.

For example, if you have a boolean variable that represents whether a user is logged in or not, you might name it "isLoggedIn" or "shouldLogin". This makes it clear that the variable represents a true or false condition related to the login state of the user.

Functions, on the other hand, are blocks of code that perform a specific action or task. Starting their names with a verb helps to indicate that they are functions and makes it clear that they are performing an action.

For example, if you have a function that adds two numbers together, you might name it "addNumbers". This makes it clear that the function performs an action and helps to distinguish it from other types of variables or objects in the code.

In general, following these naming conventions can help make your code more readable and easier to understand for other developers who may work on the same codebase in the future.

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+      break;
+
+    default:
+      break;
+  }
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
```
Andr1yK, line 59: u you can reduce the complexity and make reverse inside sort function)

src/App.tsx
```diff
@@ -15,48 +19,123 @@ export const goodsFromServer = [
...
+
+    case SortType.LENGTH:
+      visibleGoods.sort(
+        (prevGood, currGood) => prevGood.length - currGood.length,
+      );
+      break;
+
+    default:
+      break;
+  }
```
Andr1yK, line 54: DRY :)

let's use single sort

Andr1yK (changes requested):

PR #810, opened 2023-03-22
solution

Nordexz:
[DEMO LINK](https://nordexz.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 import classNames from 'classnames';
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -17,9 +18,9 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'NONE',
+  ALPHABET = 'ALPHABET',
+  LENGTH = 'LENGTH',
 }
 
 type ReorderOptions = {
@@ -30,17 +31,16 @@ type ReorderOptions = {
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
-) {
+): string[] {
   const visibleGoods = [...goods];
 
   visibleGoods.sort((g1, g2) => {
     switch (sortType) {
-      case SortType.NONE:
-        return 0;
       case SortType.ALPHABET:
         return g1.localeCompare(g2);
       case SortType.LENGTH:
         return g1.length - g2.length;
+      case SortType.NONE:
       default:
         return 0;
     }
@@ -54,11 +54,11 @@ export function getReorderedGoods(
 }
 
 export const App: React.FC = () => {
-  const [isReversed, setReversed] = useState(false);
+  const [isReversed, setIsReversed] = useState(false);
   const [sortType, setSortType] = useState(SortType.NONE);
 
   const reverse = () => {
-    setReversed(current => !current);
+    setIsReversed(current => !current);
   };
 
   const sortByType = (sortingType: SortType) => {
@@ -66,11 +66,11 @@ export const App: React.FC = () => {
   };
 
   const reset = () => {
-    setReversed(false);
+    setIsReversed(false);
     setSortType(SortType.NONE);
   };
 
-  const rstBttn = sortType !== SortType.NONE || isReversed;
+  const isResetButton = sortType !== SortType.NONE || isReversed;
   const orderingGoods = getReorderedGoods(
     goodsFromServer, { isReversed, sortType },
   );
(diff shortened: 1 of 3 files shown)


arkushenko-anna (approved):
GJ!

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
```
sergii-nosachenko, line 22: Add values to enum

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
```
sergii-nosachenko, line 33: Add return type

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((g1, g2) => {
+    switch (sortType) {
+      case SortType.NONE:
+        return 0;
```
sergii-nosachenko, line 39: Combine it with default case

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
```
sergii-nosachenko, line 57:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
+  const sortByType = (sortingType: SortType) => {
+    setSortType(sortingType);
+  };
+
+  const reset = () => {
+    setReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const rstBttn = sortType !== SortType.NONE || isReversed;
```
sergii-nosachenko, line 73: What is `rstBttn`? Name it as a normal boolean value.

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
+  const orderingGoods = getReorderedGoods(
```
sergii-nosachenko, line 74: This constant is not a process, use a proper name

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {orderingGoods
+          .map(
+            word => <li data-cy="Good" key={word}>{word}</li>,
+          )}
       </ul>
```
sergii-nosachenko, line 136: Create a separate component for the list

src/App.tsx
```diff
@@ -15,48 +16,123 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {orderingGoods
+          .map(
+            word => <li data-cy="Good" key={word}>{word}</li>,
```
sergii-nosachenko, line 134: Split into rows

sergii-nosachenko (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE = 'NONE',
+  ALPHABET = 'ALPHABET',
+  LENGTH = 'LENGTH',
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
```
Andr1yK, line 29:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+  const sortByType = (sortingType: SortType) => {
+    setSortType(sortingType);
+  };
+
+  const reset = () => {
+    setIsReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const isResetButton = sortType !== SortType.NONE || isReversed;
```
Andr1yK, line 73:
```suggestion
  const isResetButtonVisible = sortType !== SortType.NONE || isReversed;
```

src/App.tsx
Andr1yK: let's fix

Andr1yK (approved):

PR #814, opened 2023-03-23
goods hook done

Yekatery:
[DEMO LINK](https://Yekatery.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import classNames from 'classnames';
+import { GoodList } from './components/GoodsList';
+import { Button } from './components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


ksavelyeva (approved):

src/App.tsx
```diff
@@ -15,49 +19,110 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
```
Andr1yK, line 31:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,49 +19,110 @@ export const goodsFromServer = [
...
+  const reverseOrder = () => {
+    changeIsReversed(curState => !curState);
+  };
+
+  const resetSorting = () => {
+    changeIsReversed(false);
+    changeSortType(SortType.NONE);
+  };
+
+  const isChanchedOrder = sortType !== SortType.NONE || isReversed;
```
Andr1yK, line 76:
```suggestion
  const isChangedOrder = sortType !== SortType.NONE || isReversed;
```

src/components/Button.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+interface Props {
+  name: string,
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void,
+  className: string,
+}
+
+export class Button extends React.PureComponent<Props, {}> {
+  render(): React.ReactNode {
```
Andr1yK, line 10: let's use functional component

Andr1yK (changes requested):

PR #818, opened 2023-03-23
Application refactored using React Hooks

MaksymMaliuk:
[DEMO LINK](https://MaksymMaliuk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,19 +23,19 @@ enum SortType {
   LENGTH,
 }
 
-type ReorderOptions = {
-  sortType: SortType,
+interface ReorderOptions {
+  sortByType: SortType,
   isReversed: boolean,
-};
+}
 
 export function getReorderedGoods(
   goods: string[],
-  { sortType, isReversed }: ReorderOptions,
+  { sortByType, isReversed }: ReorderOptions,
 ) {
   const visibleGoods = [...goods];
 
   visibleGoods.sort((previousGood, currentGood) => {
-    switch (sortType) {
+    switch (sortByType) {
       case SortType.ALPHABET:
         return previousGood.localeCompare(currentGood);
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
```
Andr1yK, line 30:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+    }
+  });
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
```
Andr1yK, line 56: In React, when using the `useState` hook to manage state, it’s a common convention to name the state variable and its corresponding setter function using a specific naming template: `[variableName, “set” + variableName]`.

```js
const [isReversed, setReversed] = useState(false);
```

```js
```

src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+  const [sortType, setSortByType] = useState(SortType.NONE);
+  const isVisibleReset = isReversed || sortType !== SortType.NONE;
```
Andr1yK, line 58:
```suggestion
  const isResetButtonVisible = isReversed || sortType !== SortType.NONE;
```

src/App.tsx
```diff
@@ -15,49 +17,118 @@ export const goodsFromServer = [
...
+  const goods = getReorderedGoods(goodsFromServer, {
+    isReversed,
+    sortType,
+  });
+
+  const reverseList = () => {
+    setReversed(!isReversed);
```
Andr1yK, line 65: In ReactJS, it’s recommended to use functional `setState` instead of the regular `setState`. The reason for this is that the functional form of `setState` can help avoid issues with stale state, and ensure that the state is updated correctly.
```jsx
// using functional setState
```
```jsx
// using regular setState
```
```jsx
```
```jsx
this.setState({ items: [] });
```

PR #819, opened 2023-03-23
list of goods hooks solution

ksavelyeva:
[DEMO LINK](https://ksavelyeva.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import { List } from './components/List';
+import { Button } from './components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,49 +19,110 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE = 'none',
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
```
Andr1yK, line 31:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,49 +19,110 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, changeIsReversed] = useState(false);
+  const [sortType, changeSortType] = useState(SortType.NONE);
```
Andr1yK, line 61: In React, when using the `useState` hook to manage state, it’s a common convention to name the state variable and its corresponding setter function using a specific naming template: `[variableName, “set” + variableName]`.

```js
```

```js
```

src/App.tsx
```diff
@@ -15,49 +19,110 @@ export const goodsFromServer = [
...
+
+  const sortByType = (typeOfSort: SortType) => {
+    changeSortType(typeOfSort);
+  };
```
Andr1yK, line 65: let's create functions for each type or just use changeSortType cuz now sortByType just a needless wrapper

src/components/Button.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+interface Props {
+  name: string,
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void,
+  className: string,
+}
+
+export class Button extends React.PureComponent<Props, {}> {
+  render(): React.ReactNode {
```
Andr1yK, line 10: why do u need to use class component here?

Andr1yK (changes requested):

src/components/Button.tsx
```diff
@@ -0,0 +1,20 @@
...
+}) => (
+  <button
+    type="button"
+    onClick={onClick}
+    className={className}
+    {...rest}
+  >
+    {name}
+  </button>
+);
```
sergii-nosachenko, line 20: 🔥

sergii-nosachenko (approved):
Good work 👍

PR #823, opened 2023-03-23
done task with using react hooks

BozhenaBodnar:
[DEMO LINK](https://BozhenaBodnar.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { Goods } from './components/Goods';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,49 +17,120 @@ export const goodsFromServer = [
...
+      }
+    });
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
```
Andr1yK, line 55: u you can reduce the complexity and make reverse inside sort function)

src/App.tsx
```diff
@@ -15,49 +17,120 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
Andr1yK, line 122: use const for this condition

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,121 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType;
+  isReversed: boolean;
+};
```
Andr1yK, line 29: :fire:

Andr1yK:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,49 +17,120 @@ export const goodsFromServer = [
...
+  const sortByAlphabet = () => {
+    setSort(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    setSort(SortType.LENGTH);
+  };
+
+  const reverseList = () => {
+    setReverse(!isReversed);
```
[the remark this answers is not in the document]
Andr1yK: not fixed - use functional setState

Andr1yK (changes requested):

src/App.tsx
BozhenaBodnar: ops, sorry, my fault

Andr1yK: This is just a recommendation))

BozhenaBodnar: it's about not fixed state too

src/App.tsx
```diff
@@ -15,49 +17,121 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
```
Andr1yK, line 26: nit: Consider using an interface instead of a type alias for object types, as it is more flexible and easier to read. Change `type ReorderOptions` to `interface ReorderOptions`.

src/App.tsx
```diff
@@ -15,49 +17,121 @@ export const goodsFromServer = [
...
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType !== SortType.NONE) {
+    visibleGoods.sort((good1, good2) => {
```
Andr1yK, line 38: 🐞 Instead of using a magic number like `!== SortType.NONE`, consider using a more descriptive enumeration value, such as `SortType.UNSORTED`. This will improve the code readability.

Andr1yK (approved):

PR #824, opened 2023-03-23
React List of Goods (Hooks)

VadKol:
- [DEMO LINK](https://VadKol.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,63 +1,133 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { GoodsList } from './components/GoodsList';
+
+import { goodsFromServer } from './data/GoodsFromServer';
+
 import 'bulma/css/bulma.css';
+
 import './App.scss';
 
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,63 +1,133 @@
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
```
Andr1yK, line 20:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -1,63 +1,133 @@
...
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
 
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const isResetBtn = sortType !== SortType.NONE || isReversed;
```
Andr1yK, line 51: avd shrting nms

src/App.tsx
```diff
@@ -1,63 +1,133 @@
...
+    goodsFromServer,
+    {
+      sortType,
+      isReversed,
+    },
+  );
+
+  const handleSortByABC = () => setSortType(SortType.ALPHABET);
+  const handleSortByLength = () => setSortType(SortType.LENGTH);
+  const handleReverse = () => setIsReversed(!isReversed);
```
Andr1yK, line 62: In ReactJS, it’s recommended to use functional `setState` instead of the regular `setState`. The reason for this is that the functional form of `setState` can help avoid issues with stale state, and ensure that the state is updated correctly.

```jsx
// using functional setState
```

```jsx
// using regular setState
```

```jsx
```

```jsx
this.setState({ items: [] });

```

src/App.tsx
VadKol: Fixed

src/App.tsx
VadKol: Fxd

Andr1yK (approved):

PR #825, opened 2023-03-23
task completed

ArthurKrepko:
[DEMO LINK](https://ArthurKrepko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,12 +23,17 @@ enum SortType {
   LENGTH,
 }
 
+interface GoodsSortingOptions {
+  isReversed: boolean;
+  sortType: SortType;
+}
+
 export function getReorderedGoods(
   goods: string[],
-  sortType: SortType,
-  isReversed: boolean,
+  options: GoodsSortingOptions,
 ) {
   const visibleGoods = [...goods];
+  const { sortType, isReversed } = options;
 
   visibleGoods.sort((prev, curr) => {
     switch (sortType) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +17,115 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+export function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
```
Andr1yK, line 29: When working with functions that have a lot of arguments, it can sometimes be difficult to remember the order of the arguments or even to know what each argument is for. To improve the readability and maintainability of your code, it’s recommended to group related arguments together in an options object instead.

In the example provided, `isReversed` and `sortType` are two arguments that are related to the sorting of goods. Instead of passing them as separate arguments, we can group them together in an options object:

```js
export interface GoodsSortingOptions {
  isReversed: boolean;
  sortType: SortType;
}

  goods: string[],
  options: GoodsSortingOptions
) {
...
}
```

This approach has several advantages:

- **Readability**: It’s easier to understand what each argument is for when they are grouped together in an options object.

- **Flexibility**: We can add new options without breaking existing code. For example, if we want to add a `maxItems` option, we can simply add it to the `GoodsSortingOptions` interface.

- **Reduced cognitive load**: When calling the function, we only need to pass in a single argument instead of multiple arguments.

Here’s an example of how we can call the function with the new options object:

```js
const goods = [“apple”, “banana”, “orange”];
const options = {
  isReversed: true,
  sortType: SortType.ASCENDING,
};

const reorderedGoods = getReorderedGoods(goods, options);
```

In summary, grouping related arguments together in an options object can improve the readability and maintainability of your code, and reduce the cognitive load of calling the function.

src/App.tsx
```diff
@@ -15,49 +17,115 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, sortByType] = useState(SortType.NONE);
+  const [isReversed, setReverse] = useState(false);
```
Andr1yK, line 53: In React, when using the `useState` hook to manage state, it’s a common convention to name the state variable and its corresponding setter function using a specific naming template: `[variableName, “set” + variableName]`.

```js
```

```js
```

src/App.tsx
```diff
@@ -15,49 +17,115 @@ export const goodsFromServer = [
...
+    sortByType(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    sortByType(SortType.LENGTH);
+  };
+
+  const reverse = () => {
+    setReverse(!isReversed);
+  };
```
Andr1yK, line 65: In ReactJS, it’s recommended to use functional `setState` instead of the regular `setState`. The reason for this is that the functional form of `setState` can help avoid issues with stale state, and ensure that the state is updated correctly.

```jsx
// using functional setState
```

```jsx
// using regular setState
```

```jsx
```

```jsx
this.setState({ items: [] });

```

src/App.tsx
```diff
@@ -15,49 +17,115 @@ export const goodsFromServer = [
...
+                onClick={reset}
+              >
+                Reset
+              </button>
+            )}
+        </div>
+
+        <ListOfGoods goods={reorderedGoods} />
+      </div>
+    </>
```
Andr1yK, line 129: why do u need this fragment?

src/App.tsx
```diff
@@ -15,49 +17,115 @@ export const goodsFromServer = [
...
+            className={classNames(
+              'button is-warning',
+              { 'is-light': isReversed === false },
+            )}
+            onClick={reverse}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== SortType.NONE || isReversed)
```
Andr1yK, line 115: let's move this condition to const

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,122 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
```
sergii-nosachenko, line 23: Always add values to enum

src/App.tsx
```diff
@@ -15,49 +17,122 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReverse] = useState(false);
```
sergii-nosachenko, line 58:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,49 +17,122 @@ export const goodsFromServer = [
...
+
+  const sortByAlphabet = () => {
+    setSortType(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const reverse = () => {
```
sergii-nosachenko, line 68:
```suggestion
  const reverseOrder = () => {
```

src/App.tsx
```diff
@@ -15,49 +17,122 @@ export const goodsFromServer = [
...
+    setReverse(prevReverse => !prevReverse);
+  };
+
+  const reset = () => {
```
sergii-nosachenko, line 72:
```suggestion
  const resetSorting = () => {
```

src/App.tsx
```diff
@@ -15,49 +17,122 @@ export const goodsFromServer = [
...
+
+  const reorderedGoods = getReorderedGoods(
+    goodsFromServer,
+    {
+      sortType,
+      isReversed,
+    },
+  );
+
+  const noneSorted = sortType !== SortType.NONE;
```
sergii-nosachenko, line 85: `isSorted`

PR #826, opened 2023-03-24
Develop

arkushenko-anna:
[DEMO LINK](https://arkushenko-anna.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
```
Andr1yK, line 53: u you can reduce the complexity and make reverse inside sort function)

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+  const visibitityResetButton = (sortType !== SortType.NONE || isReversed);
```
Andr1yK, line 58: Naming conventions for boolean variables and functions are important because they help make code more readable and easier to understand.

Boolean variables, which represent true or false values, are often used to store the state of a condition in a program. Starting their names with "is", "should", or "does" helps to indicate that they are boolean variables and makes it clear that their value represents a true or false condition.

For example, if you have a boolean variable that represents whether a user is logged in or not, you might name it "isLoggedIn" or "shouldLogin". This makes it clear that the variable represents a true or false condition related to the login state of the user.

Functions, on the other hand, are blocks of code that perform a specific action or task. Starting their names with a verb helps to indicate that they are functions and makes it clear that they are performing an action.

For example, if you have a function that adds two numbers together, you might name it "addNumbers". This makes it clear that the function performs an action and helps to distinguish it from other types of variables or objects in the code.

In general, following these naming conventions can help make your code more readable and easier to understand for other developers who may work on the same codebase in the future.

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+interface Props {
+  goods: string[]
+}
+
+export const GoodsList: React.FC<Props> = ({ goods }) => {
+  return (
```
Andr1yK, line 8: useless return

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <ul>
+      {
+        goods.map(good => (
```
Andr1yK, line 11:
```suggestion
      {goods.map(good => (
```

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
```
Andr1yK, line 1: nit: Import useState directly `import { useState } from 'react';`. No need to import React since you are not using it directly 🧐

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames(
+            'button is-inf',
```
Andr1yK, line 78: nit: Typo in class name, it should be 'button is-info'. Fix 'button is-inf' 🚀

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((prevGood, currentGood) => {
```
Andr1yK, line 37: nit: Use more explicit variable names for prevGood and currentGood, such as prevItem and currentItem, for better readability 📘

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
```
Andr1yK, line 26: nit: Consider replacing 'ReorderOptions' with a more meaningful name like 'SortOptions' to match the sorting context 💡

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
```
Andr1yK, line 23: let's add values)

NONE = "NONE"

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        { isResetButtonVisible && (
```
Andr1yK, line 108:
```suggestion
        {isResetButtonVisible && (
```

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,106 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
```
Andr1yK, line 57: nit: Use a more semantic variable name, e.g., isSortingReversed instead of isReversed. This will improve code readability. 🎓

PR #828, opened 2023-03-24
Solution hooks

nazarovoleksiy:
[Demo](https://nazarovoleksiy.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,124 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType;
+  isReversed: boolean;
+};
```
Andr1yK, line 29:
```typescript
interface Person {
  name: string;
  age: number;
}

```

src/App.tsx
```diff
@@ -15,48 +17,124 @@ export const goodsFromServer = [
...
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((goodOne, goodTwo) => {
```
Andr1yK, line 37:
```suggestion
  visibleGoods.sort((goodFirst, goodSecond) => {
```

Andr1yK (approved):

PR #856, opened 2023-04-13
created a goods func with hooks

prokivskaaa:
[DEMO LINK](https://prokivskaaa.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,21 +33,18 @@ export function getReorderedGoods(
 ) {
   const visibleGoods = [...goods];
 
-  const isAlphabeticSort = sortType === SortType.ALPHABET;
-  const isLengthSort = sortType === SortType.LENGTH;
+  visibleGoods.sort((firstGood, secondGood) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return firstGood.localeCompare(secondGood);
 
-  switch (true) {
-    case isAlphabeticSort:
-      visibleGoods.sort((g1, g2) => g1.localeCompare(g2));
-      break;
+      case SortType.LENGTH:
+        return firstGood.length - secondGood.length;
 
-    case isLengthSort:
-      visibleGoods.sort((g1, g2) => g1.length - g2.length);
-      break;
-
-    default:
-      break;
-  }
+      default:
+        return 0;
+    }
+  });
 
   if (isReversed) {
     visibleGoods.reverse();
@@ -75,7 +72,7 @@ export const App: React.FC = () => {
   };
 
   const handleReverse = () => {
-    setIsReversed((currentReversed) => !currentReversed);
+    setIsReversed(!isReversed);
   };
 
   const handleSort = (type: SortType) => {


src/App.tsx
```diff
@@ -15,48 +16,117 @@ export const goodsFromServer = [
...
+    setSortType(SortType.NONE);
+  };
+
+  const reverseHandler = () => {
+    setIsReversed((currentReversed) => !currentReversed);
+  };
+
+  const sortHandler = (type: SortType) => {
+    setSortType(type);
+  };
```
Esteban4769, line 78: Not really right way to name handlers. Check [checklist](https://github.com/mate-academy/react_list-of-goods-hooks/blob/master/checklist.md) pt.4

src/App.tsx
```diff
@@ -15,48 +16,117 @@ export const goodsFromServer = [
...
+  const isAlphabeticSort = sortType === SortType.ALPHABET;
+  const isLengthSort = sortType === SortType.LENGTH;
+
+  if (isAlphabeticSort) {
+    visibleGoods.sort((g1, g2) => g1.localeCompare(g2));
+  }
+
+  if (isLengthSort) {
+    visibleGoods.sort((g1, g2) => g1.length - g2.length);
+  }
```
Esteban4769, line 45: I would probably use `switch-case` for `sortType`, but its up to you.

Esteban4769 (approved):
Good work, can make some adjustment tho

src/App.tsx
jstmpelowycz: agree, let's apply switch-case, it's right the case to use it, put it inside compare callback

src/App.tsx
```diff
@@ -15,48 +16,117 @@ export const goodsFromServer = [
...
+  const isResetButtonShown = sortType !== SortType.NONE || isReversed;
+  const isAlphabeticSort = sortType === SortType.ALPHABET;
+  const isLengthSort = sortType === SortType.LENGTH;
+
+  const goods = getReorderedGoods(goodsFromServer, {
+    sortType,
+    isReversed,
+  });
+
+  const resetHandler = () => {
```
jstmpelowycz, line 67:
```js
const '<do><what>' = ...;
```
```suggestion
  const handleReset = () => {
```

jstmpelowycz (changes requested):

src/App.tsx
prokivskaaa: @jstmpelowycz Thank you! Fixed)

src/App.tsx
```diff
@@ -15,48 +16,122 @@ export const goodsFromServer = [
...
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  const isAlphabeticSort = sortType === SortType.ALPHABET;
+  const isLengthSort = sortType === SortType.LENGTH;
+
+  switch (true) {
+    case isAlphabeticSort:
```
jstmpelowycz, line 40:
```suggestion
  switch (sortType) {
    case SortType.ALPHABET:
```
and so on

jstmpelowycz: > The switch statement evaluates an expression. The value of the expression is then compared with the values of each case in the structure. If there is a match, the associated block of code is executed.

src/App.tsx
```diff
@@ -15,48 +16,122 @@ export const goodsFromServer = [
...
+    isReversed,
+  });
+
+  const handleReset = () => {
+    setIsReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const handleReverse = () => {
+    setIsReversed((currentReversed) => !currentReversed);
```
jstmpelowycz, line 78:
```suggestion
    setIsReversed(!isReversed);
```

src/App.tsx
jstmpelowycz: also, use switch inside sorting callback
as you can see, you repeat yourself
```js
visibleGoods.sort(...);
...

visibleGoods.sort(...);
...

**Review by jstmpelowycz — CHANGES_REQUESTED** [review]

File: src/App.tsx (hunk 5 above)
Comment (prokivskaaa) [reply src/App.tsx]: @jstmpelowycz Thank you! Done)

File: src/App.tsx [hunk 6]
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
+    setSortType(type);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button', 'is-info', {
```
SerhiiKirik, line 87: you can just combine it
```suggestion
          className={classNames('button is-info', {
```

SerhiiKirik (approved):
Great 🔥

PR #857, opened 2023-04-13
add task solution

Peti51:
[DEMO LINK](https://Peti51.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,26 +26,23 @@ function getReorderedGoods(
   isReversed: boolean,
   sortType: SortType,
 ):string[] {
-  const visible = [...goods];
+  const visibleGoods = [...goods];
 
-  visible.sort((goodFirst, goodSecond) => {
+  visibleGoods.sort((goodFirst, goodSecond) => {
     switch (sortType) {
       case SortType.ALPHABET:
         return goodFirst.localeCompare(goodSecond);
       case SortType.LENGTH:
         return goodFirst.length - goodSecond.length;
       case SortType.NONE:
-        return 0;
       default:
         return 0;
     }
   });
 
-  if (isReversed) {
-    visible.reverse();
-  }
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
 
-  return visible;
+  return visibleGoods;
 }
 
 export const App: React.FC = () => {
@@ -57,9 +54,16 @@ export const App: React.FC = () => {
     setIsReversed(prevState => !prevState);
   };
 
-  const changedArr: string[]
-  = useMemo(() => getReorderedGoods(copiedData, isReversed, sortType),
-    [copiedData, isReversed, sortType]);
+  const onReset = () => {
+    setIsReversed(false);
+
+    setSortType(SortType.NONE);
+  };
+
+  const changedArr: string[] = useMemo(
+    () => getReorderedGoods(copiedData, isReversed, sortType),
+    [copiedData, isReversed, sortType],
+  );
 
   return (
     <div className="section content">
@@ -88,20 +92,15 @@ export const App: React.FC = () => {
           Reverse
         </button>
 
-        {isReversed || sortType !== SortType.NONE
-          ? (
-            <button
-              type="button"
-              className="button is-danger is-light"
-              onClick={() => {
-                setIsReversed(false);
-                setSortType(SortType.NONE);
-              }}
-            >
-              Reset
-            </button>
-          )
-          : null}
+        {(isReversed || sortType !== SortType.NONE) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={onReset}
+          >
+            Reset
+          </button>
+        )}
       </div>
 
       <ul>


src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+    visible.reverse();
+  }
+
+  return visible;
+}
+
 export const App: React.FC = () => {
+  const copiedData = [...goodsFromServer];
+  const [trueOrFalseReverse, reversing] = useState(false);
+  const [sortType, sorting] = useState(SortType.NONE);
```
qoa11a, line 54:
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+
+  const reverse = () => {
+    reversing(current => !current);
```
qoa11a, line 57:
```suggestion
    setIsReversed(prevState => !prevState);
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+          onClick={() => sorting(SortType.ALPHABET)}
         >
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={`button is-success ${sortType === SortType.LENGTH ? '' : 'is-light'}`}
+          onClick={() => sorting(SortType.LENGTH)}
```
qoa11a, line 78: Better to move this to a separate function. Like with reverse.

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {changedArr.map((good) => {
+            return <li data-cy="Good" key={good}>{good}</li>;
+          })}
```
qoa11a, line 111:
```suggestion
          {changedArr.map((good) => (
            <li data-cy="Good" key={good}>{good}</li>;
          ))}
```

qoa11a (changes requested):

Peti51:
What do you mean by that mov to a function like with reverse?

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+  visible.sort((goodFirst, goodSecond) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return goodFirst.localeCompare(goodSecond);
+      case SortType.LENGTH:
+        return goodFirst.length - goodSecond.length;
+      case SortType.NONE:
+        return 0;
+      default:
+        return 0;
```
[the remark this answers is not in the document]
Anastasiia-Svintsova: Not fixed. You can combine cases with similar actions
```suggestion
      case SortType.NONE:
      default:
        return 0;
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+  ALPHABET = 1,
+  LENGTH = 2,
+}
+
+function getReorderedGoods(
+  goods: string[],
+  isReversed: boolean,
+  sortType: SortType,
+):string[] {
+  const visible = [...goods];
```
Anastasiia-Svintsova, line 29: use better naming for the variables 
```suggestion
  const visibleGoods = [...goods];
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visible.reverse();
+  }
+
+  return visible;
```
Anastasiia-Svintsova, line 48: you could simplify it 
```suggestion
  return isReversed ? visible.reverse() : visible;
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const reverse = () => {
+    setIsReversed(prevState => !prevState);
+  };
+
+  const changedArr: string[]
+  = useMemo(() => getReorderedGoods(copiedData, isReversed, sortType),
+    [copiedData, isReversed, sortType]);
```
Anastasiia-Svintsova, line 62: Check the code style, It's not easy to read it now. BTW you don't really need to specify the variable type here.
```suggestion
  const changedArr: string[] = useMemo(
    () => getReorderedGoods(copiedData, isReversed, sortType),
    [copiedData, isReversed, sortType]
  );
```

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
+              className="button is-danger is-light"
+              onClick={() => {
+                setIsReversed(false);
+                setSortType(SortType.NONE);
+              }}
+            >
+              Reset
+            </button>
+          )
+          : null}
```
Anastasiia-Svintsova, line 104: You don't need to use a ternary operator here. Just use `&&` If you need to render nothing if the condition is not true

src/App.tsx
```diff
@@ -15,47 +15,100 @@ export const goodsFromServer = [
...
-        </button>
+        {isReversed || sortType !== SortType.NONE
+          ? (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => {
+                setIsReversed(false);
+                setSortType(SortType.NONE);
+              }}
```
Anastasiia-Svintsova, line 99: Create a function and move the logic there 
```suggestion
              onClick={onReset}
```

Anastasiia-Svintsova (changes requested):

src/App.tsx
```diff
@@ -15,47 +15,99 @@ export const goodsFromServer = [
...
+        return goodFirst.length - goodSecond.length;
+      case SortType.NONE:
+      default:
+        return 0;
+    }
+  });
+
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
+
+  return visibleGoods;
```
polosanya, line 45:
```suggestion
  return isReversed ? visibleGoods.reverse() : visibleGoods;
```
Second return is redundant. You will return the goods anyway with ternary operator

src/App.tsx
```diff
@@ -15,47 +15,99 @@ export const goodsFromServer = [
...
 export const App: React.FC = () => {
+  const copiedData = [...goodsFromServer];
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const reverse = () => {
+    setIsReversed(prevState => !prevState);
+  };
+
+  const onReset = () => {
```
polosanya, line 57:
```suggestion
  const handleReset = () => {
```
https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/App.tsx
```diff
@@ -15,47 +15,99 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
polosanya, line 95: It would be great to create a separate variable for this condition

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #858, opened 2023-04-14
solution

ladrom:
[demo](https://ladrom.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,47 +16,140 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
```
Volodymyr-Melnyk-Fron10der, line 20:
```suggestion
  None,
```

fix other properties in the enum as well

Volodymyr-Melnyk-Fron10der (approved):
Approve, but keep in mind that it's better to capitalize enum properties rather then uppercase them

src/App.tsx
ladrom: Enums were originally in the uppercase from the last task, but thanks for the feedback.

PR #864, opened 2023-04-18
solution

PashaMalyshkin:
[DEMO LINK](https://PashaMalyshkin.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +15,88 @@ export const goodsFromServer = [
...
+  const visibleGoods = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={sortType === SortType.ALPHABET
+            ? 'button is-info' : 'button is-info is-light'}
```
igorkruz, line 58: I'm not sure but I think in all cases like that it's better to use `classNames`

src/App.tsx
```diff
@@ -15,48 +15,88 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed) && (
```
igorkruz, line 82: Probably it would be better if you create a variable for this expression.
It would be more readable

igorkruz (approved):
Good job 👍 
 You can make some adjustments too.

src/App.tsx
```diff
@@ -15,48 +15,88 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+export function getReorderedGoods(
+  goods: string[], sortType : SortType, isReversed : boolean,
+) {
```
jstmpelowycz, line 26:
```suggestion
export function getReorderedGoods(
  goods: string[], 
  sortType: SortType, 
  isReversed: boolean,
) {
```

src/App.tsx
```diff
@@ -15,48 +15,88 @@ export const goodsFromServer = [
...
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((firstGood, secondGood) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return firstGood.localeCompare(secondGood);
+      case SortType.LENGTH:
+        return firstGood.length - secondGood.length;
+      default:
+        return 0;
```
jstmpelowycz, line 36:
```suggestion
      case SortType.NONE:
      default:
        return 0;
```

src/App.tsx
jstmpelowycz: agree, use `classnames`
here and everywhere else

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,97 @@ export const goodsFromServer = [
...
+
+  visibleGoods.sort((firstGood, secondGood) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return firstGood.localeCompare(secondGood);
+      case SortType.LENGTH:
+        return firstGood.length - secondGood.length;
+      case SortType.NONE:
+      default:
+        return 0;
```
jstmpelowycz, line 40:
```suggestion
    switch (sortType) {
      case SortType.ALPHABET:
        return firstGood.localeCompare(secondGood);
        
      case SortType.LENGTH:
        return firstGood.length - secondGood.length;
        
      case SortType.NONE:
      default:
        return 0;
```

src/App.tsx
```diff
@@ -15,48 +16,97 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames({
+            'button is-info is-light': sortType !== SortType.ALPHABET,
+            'button is-info': sortType === SortType.ALPHABET,
+          })}
```
jstmpelowycz, line 64:
```suggestion
          className={cn('button is-info', { 
            ' is-light': sortType !== SortType.ALPHABET,
          })}
```
and so on

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,94 @@ export const goodsFromServer = [
...
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, sortBy] = useState(SortType.NONE);
+  const [isReversed, reverseGoods] = useState(false);
+  const visibleGoods = getReorderedGoods(goodsFromServer, sortType, isReversed);
```
jstmpelowycz, line 54:
```suggestion
  const [sortType, sortBy] = useState(SortType.NONE);
  const [isReversed, reverseGoods] = useState(false);

  const visibleGoods = getReorderedGoods(
    goodsFromServer, 
    sortType, 
    isReversed,
  );
```

src/App.tsx
```diff
@@ -15,48 +16,94 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map(good => (
+          <li data-cy="Good" key={good}>{good}</li>
```
jstmpelowycz, line 105:
```suggestion
          <li data-cy="Good" key={good}>
            {good}
          </li>
```

jstmpelowycz (approved):

PR #870, opened 2023-04-20
The task has been done using hooks

bojkovladislav:


src/App.tsx
```diff
@@ -16,46 +16,90 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState('none');
```
anstsot, line 19: please create `enum SortType` and use it everywhere in the component

src/App.tsx
```diff
@@ -16,46 +16,90 @@ export const goodsFromServer = [
...
+  const [isReversed, setDirection] = useState(false);
+
+  const reset = () => {
```
anstsot, line 22: let's rename the function according to standards: `on` + `Action`.
please check all cases
```suggestion
  const onReset = () => {
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/App.tsx
```diff
@@ -15,47 +15,97 @@ export const goodsFromServer = [
...
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE
+          || isReversed)
+          && (
```
loralevitska, line 91:
```suggestion
        {(sortType !== SortType.NONE || isReversed) && (
```

loralevitska (approved):
Good job!

PR #875, opened 2023-04-21
list of goods hooks is solved

M1k1ta:
[DEMO LINK](https://m1k1ta.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,104 @@ export const goodsFromServer = [
...
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
KosolapovAnatolii, line 52: Why you saved it? I think you can delete it

src/App.tsx
```diff
@@ -15,48 +16,104 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.ALPHABET,
+          })}
+          onClick={() => setSortType(SortType.ALPHABET)}
```
KosolapovAnatolii, line 72: You also can use the separate function for this and another buttons.
```suggestion
          onClick={sortByAlphabet}
```

KosolapovAnatolii (changes requested):
Looking very good!
Let's make some small improvments

SerhiiKirik (approved):
Well done 🚀

PR #882, opened 2023-04-21
implementation list of goods use hooks

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { FC, useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,110 @@ export const goodsFromServer = [
...
+  const isSelected = isReversed || sortType !== SortType.NONE;
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.ALPHABET,
```
Nazarii98, line 78: You should destruct here.
Also in the next cases under this.

Nazarii98 (changes requested):
Good job!

I see one moment which you should simplify.

PR #891, opened 2023-04-22
Solution

vksazonov:
[DEMO LINK](https://vksazonov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { FC, useState } from 'react';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,48 +15,109 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.ALPHABET,
+          })}
+          onClick={handleAlphabetSort}
```
AlexRog01, line 80: Here it was possible to enter at once a function with the transferred necessary data. Then you wouldn't have to write so many handlers

AlexRog01 (approved):
Good

jstmpelowycz (approved):

PR #900, opened 2023-04-23
implementation of the task using hooks

TolikSerhiychuk:
[DEMO LINK](https://TolikSerhiychuk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +15,115 @@ export const goodsFromServer = [
...
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((good1, good2) => good1.length - good2.length);
+  }
```
vksazonov, line 41: Maybe it's better to use switch over here

vksazonov (approved):
Good Job!

src/App.tsx
```diff
@@ -15,49 +15,115 @@ export const goodsFromServer = [
...
+  const handleAlphabetSort = () => {
+    setSortType(SortType.ALPHABET);
+  };
+
+  const handleLengthSort = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const handleReverse = () => {
+    setIsReversed(currentIsReversed => !currentIsReversed);
```
jstmpelowycz, line 63: you already have access to current `isReversed`
```suggestion
    setIsReversed(!isReversed);
```

src/App.tsx
```diff
@@ -15,49 +15,115 @@ export const goodsFromServer = [
...
+  };
+
+  const handleReset = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
+
+  const goods = getReorderedGoods(goodsFromServer, { isReversed, sortType });
```
jstmpelowycz, line 71:
```suggestion
  const reordered = getReorderedGoods(goodsFromServer, {
    isReversed, 
    sortType,
  });
```

src/App.tsx
```diff
@@ -15,49 +15,115 @@ export const goodsFromServer = [
...
-        </button>
-      </div>
+    <>
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={sortType === SortType.ALPHABET
+              ? 'button is-info'
+              : 'button is-info is-light'}
```
jstmpelowycz, line 81: let's use classnames, everywhere
```js
import cn from 'classnames';
```

jstmpelowycz (changes requested):

src/App.tsx
TolikSerhiychuk: fixed

src/App.tsx
TolikSerhiychuk: got it

src/App.tsx
TolikSerhiychuk: fixed

jstmpelowycz (approved):

PR #905, opened 2023-04-23
Develop

IvanTkachenko3301:
[DEMO LINK](https://IvanTkachenko3301.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+const goodsFromServer = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
+  };
+
+  const reverse = () => {
+    setIsReversed(prevState => !prevState);
+  };
+
+  const reset = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
```
Stambuliak, line 73: u can create separate function with switch/case for this.

src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
+    setSortType(SortType.ALPHABET);
+  };
+
+  const sortByLength = () => {
+    setSortType(SortType.LENGTH);
+  };
+
+  const reverse = () => {
+    setIsReversed(prevState => !prevState);
+  };
```
Stambuliak, line 68: !prevState also works

Stambuliak (approved):
Looks good but u can optimize it

src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed) && (
+          <button
+            type="button"
+            className={classNames('button is-danger', {
```
AP-Solution, line 113: You can use short syntax here, like this:

className={cn('button is-danger', {

src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
+          })}
+          onClick={sortByAlphabet}
         >
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames('button is-success', {
```
AP-Solution, line 92: Short syntax can be used here, 'cn' instead of 'classNames'

src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
+
+  visibleGoods.sort((firstGood, secondGood) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return firstGood.localeCompare(secondGood);
+
+      case SortType.LENGTH:
+        return firstGood.length - secondGood.length;
+
+      default:
```
jstmpelowycz, line 42:
```suggestion
      case SortType.None:
      default:
```

src/App.tsx
jstmpelowycz: agree, you already have access to current `isReversed`

src/App.tsx
```diff
@@ -15,49 +16,117 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
+export function App() {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const sortByAlphabet = () => {
```
jstmpelowycz, line 58:
```suggestion
  const handleSortByAlphabet = () => {
```
and so on

src/App.tsx
jstmpelowycz: great idea, keep it short, it's common way to name it this way

jstmpelowycz (changes requested):

PR #926, opened 2023-05-04
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  reverse: boolean,
+  type: number,
+};
+
+function ReorderGoods(goods: string[], { reverse, type }: ReorderOptions) {
```
polosanya, line 29:
```suggestion
function getReorderGoods(goods: string[], { reverse, type }: ReorderOptions) {
```
Let's name functions according to these rules:
image

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  const initialGoods = [...goods];
+
+  switch (type) {
+    case 1:
```
polosanya, line 33: Don't hardcode the types with numbers. Use SortType enum instead

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortType, useType] = useState(SortType.NONE);
+  const [reverse, useReverse] = useState(false);
+
+  function ChangeType(type: SortType) {
+    useType(type);
+  }
+
+  function ChangeReverse(reverseType: boolean) {
```
polosanya, line 60: Function names should be in camelCase, not PascalCase

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  if (reverse) {
+    initialGoods.reverse();
+  }
+
+  return initialGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, useType] = useState(SortType.NONE);
+  const [reverse, useReverse] = useState(false);
```
polosanya, line 54:
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
  const [isReversed, setIsReversed] = useState(false);
```
Pay attention to the naming
https://react.dev/reference/react/useState

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(reverse || sortType !== SortType.NONE) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              ChangeReverse(false);
+              ChangeType(0);
+            }}
```
polosanya, line 98: Create a separate function to handle this click

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(reverse || sortType !== SortType.NONE) && (
```
polosanya, line 91: It would be great to save these conditions into some variable

polosanya (changes requested):
Good job! Let's improve your solution a bit)

PR #929, opened 2023-05-08
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+
+  if (reverse) {
+    initialGoods.reverse();
+  }
+
+  return initialGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, useType] = useState(SortType.NONE);
```
YegorKorenkov, line 53: the function that changes the state must have prefix 'set', for example setType

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  const [reverse, useReverse] = useState(false);
```
YegorKorenkov, line 54: Same

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  reverse: boolean,
+  type: number,
+};
+
+function ReorderGoods(goods: string[], { reverse, type }: ReorderOptions) {
```
YegorKorenkov, line 29: it is more correct to name the function with a start small letter

src/App.tsx
```diff
@@ -15,48 +15,102 @@ export const goodsFromServer = [
...
+  const initialGoods = [...goods];
+
+  switch (type) {
+    case 1:
```
YegorKorenkov, line 33: SortType.ALPHABET, please change for all cases

YegorKorenkov (changes requested):

PR #934, opened 2023-05-09
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +15,96 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  function reset() {
```
Radoslaw-Czerniawski, line 52: Could be `handleReset` to signify that it's a callback function for a onClick event on a button

src/App.tsx
```diff
@@ -15,47 +15,96 @@ export const goodsFromServer = [
...
+          onClick={() => setSortType(SortType.LENGTH)}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`button is-warning ${!isReversed ? 'is-light' : ''}`}
+          onClick={() => setIsReversed(state => !state)}
```
Radoslaw-Czerniawski, line 83: When handling previous state use something like `prevState` for the state parameter to signify you are working on the previous state

src/App.tsx
```diff
@@ -15,47 +15,96 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
```
Radoslaw-Czerniawski, line 49: You could have similarly used the object as a state like in the class implementation, however this if fine too

Radoslaw-Czerniawski (approved):
Left some small comments, remember to rename the previous state apart from that very nice :D

PR #957, opened 2023-05-16
Solution V1

marshalek37:
[DEMO LINK](https://marshalek37.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import React, { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,141 @@ export const goodsFromServer = [
...
+    );
+  } else if (sortType === SortType.LENGTH) {
+    visibleGoods.sort(
+      (goodA, goodB) => goodA.length - goodB.length,
+    );
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
kasiakubisiak, line 48: lets do switch/case instead of if statements

src/App.tsx
```diff
@@ -15,47 +16,141 @@ export const goodsFromServer = [
...
+  sortType: SortType,
+};
+
 export const App: React.FC = () => {
+  const [state, setState] = useState<State>({
+    isReversed: false,
+    sortType: SortType.NONE,
+  });
+
+  const sortAlphabet = () => {
```
kasiakubisiak, line 64:
```suggestion
  const sortAlphabetically = () => {
```

src/App.tsx
```diff
@@ -15,47 +16,141 @@ export const goodsFromServer = [
...
+    setState({
+      ...state,
+      sortType: SortType.ALPHABET,
+    });
+  };
+
+  const sortLength = () => {
```
kasiakubisiak, line 71:
```suggestion
  const sortByLength = () => {
```

src/App.tsx
kasiakubisiak: also, use better naming instead of goodA, gooB

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #968, opened 2023-05-20
Solution-1

MykolaButylkov:
[DEMO LINK](https://MykolaButylkov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,119 @@ export const goodsFromServer = [
...
+      visibleGoods = visibleGoods.sort(
+        (a, b) => (a.length - b.length),
+      );
+      isResetButton = true;
+      break;
+
+    case SortType.RESET:
+      visibleGoods = [...goodsFromServer];
+      isResetButton = false;
+      break;
```
VitaliyBondarenko1982, line 56:
```suggestion
    case SortType.RESET:
      visibleGoods = [...goodsFromServer];
      isResetButton = false;
      break;
```
no need to make SortType Reset and  variable isResetButton to change after each change SortType;

just make function for reset button which set SortType.None and isReversed in false.
and count isResetButton something like that one time
const isResetButton = isReversed || sortType !== SortType.NONE

src/App.tsx
```diff
@@ -15,47 +16,119 @@ export const goodsFromServer = [
...
+  NONE,
+  ALPHABET,
+  LENGTH,
+  RESET,
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortBy] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+  let isResetButton = false;
```
VitaliyBondarenko1982, line 29:
```suggestion
```

VitaliyBondarenko1982 (changes requested):

src/App.tsx
MykolaButylkov: const isResetButton = isReversed || sortType !== SortType.NONE 
 if SortType === Reset => isResetButton should be invisible, so I made another function : 
 const isResetButton = isReversed
  || sortType === SortType.ALPHABET
  || sortType === SortType.LENGTH;

src/App.tsx
```diff
@@ -15,47 +16,112 @@ export const goodsFromServer = [
...
+
+  let visibleGoods = [...goodsFromServer];
+
+  switch (sortType) {
+    case SortType.NONE:
+      break;
+
+    case SortType.ALPHABET:
+      visibleGoods = visibleGoods.sort(
+        (a, b) => (a.localeCompare(b.toString())
```
etojeDenys, line 38: Missing closing parenthesis after `a.localeCompare(b` in the sort function, should be `a.localeCompare(b)`

etojeDenys (approved):
lgtm

PR #977, opened 2023-05-29
react-list-hooks

theFallcon:
[DEMO LINK](https://theFallcon.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -16,46 +54,73 @@ export const goodsFromServer = [
...
+    goodsFromServer, { sortType, isReversed },
+  );
+
+  const hasOrderChanged = isReversed || sortType !== SortType.NONE;
+
+  const setClass = (condition: boolean) => {
+    return classNames('button', { 'is-light': condition });
+  };
+
+  const setReset = () => {
```
agniya-i, line 70:
```suggestion
  const reset = () => {
```
It would be better to remove 'set' from simple handler name to not confuse with state

agniya-i (approved):

PR #980, opened 2023-06-02
solution

devWizHub:
[DEMO LINK] https://devWizHub.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


devWizHub:
The error of not showing goods in original order after clicking the reverse button twice can not be observed, please assist. Thank you

src/App.tsx
```diff
@@ -15,47 +15,95 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type SortType = 'NONE' | 'ALPHABET' | 'LENGTH';
```
Anastasiia-Svintsova, line 18: Use an enum instead

src/App.tsx
```diff
@@ -15,47 +15,95 @@ export const goodsFromServer = [
...
+type IsReversed = boolean;
+type IsReset = boolean;
```
Anastasiia-Svintsova, line 20: You don't need to save a boolean type in a variable. Just use `boolean` directly

src/App.tsx
```diff
@@ -15,47 +15,95 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState([...goodsFromServer]);
+  const [sortType, setSortType] = useState<SortType>('NONE');
+  const [isReversed, setReverse] = useState<IsReversed>(false);
+  const [isReset, setReset] = useState<IsReset>(false);
```
Anastasiia-Svintsova, line 26: TypeScript can infer the types of the state variables isReversed and isReset based on the initial values false. Since the initial values are booleans, TypeScript will infer the types as boolean automatically.
```suggestion
  const [isReversed, setReverse] = useState(false);
  const [isReset, setReset] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +15,95 @@ export const goodsFromServer = [
...
+
+  const isOriginalOrder = !isReversed && sortType === 'NONE';
```
Anastasiia-Svintsova, line 28: Here is the reason for the issue. Try this: 
```suggestion
  const isOriginalOrder = !isReversed || sortType === 'NONE';
```

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (changes requested):
Feel free to ask questions in the chat if you need to

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+
+    switch (sortType) {
+      case 'ALPHABET':
+        sortedGoods.sort((a, b) => a.localeCompare(b));
+        break;
+      case 'LENGTH':
+        sortedGoods.sort((a, b) => {
+          return a.length !== b.length
+            ? a.length - b.length
+            : a.localeCompare(b);
```
Anastasiia-Svintsova, line 41: You don't need to use localeCompare for the length

Anastasiia-Svintsova (approved):

PR #985, opened 2023-06-05
list of good with hooks was created

Pavlo-Petrashevskyi:
[DEMO LINK](https://Pavlo-Petrashevskyi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -16,12 +16,56 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortedBy, setSortedBy] = useState('none');
+  const [isVisible, setIsVisible] = useState(false);
+
+  const changeArray = (array: string[]) => {
+    const visibleGoods = [...array];
+
+    switch (sortedBy) {
+      case 'none':
+        break;
+
+      case 'alphabet':
+        visibleGoods.sort((good1, good2) => {
+          return good1.localeCompare(good2);
+        });
+        break;
+
+      case 'length':
+        visibleGoods.sort((good1, good2) => {
+          return good1.length - good2.length;
+        });
+        break;
+
+      default:
+        break;
+    }
+
+    if (isReversed) {
+      visibleGoods.reverse();
+    }
+
+    return visibleGoods;
+  };
+
+  const newGoods = changeArray(goodsFromServer);
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
           className="button is-info is-light"
+          onClick={(e) => {
+            setSortedBy('alphabet');
+            setIsVisible(true);
+            e.currentTarget.classList.remove('is-light');
+
+            return document
+              .querySelector('.is-success')?.classList.add('is-light');
+          }}
         >
           Sort alphabetically
         </button>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,46 +16,120 @@ export const goodsFromServer = [
...
+  const [isVisible, setIsVisible] = useState(false);
+
+  const changeArray = (array: string[]) => {
+    const visibleGoods = [...array];
+
+    switch (sortedBy) {
+      case 'none':
+        break;
+
+      case 'alphabet':
```
polosanya, line 30: Don't hardcode the cases, create Enum for them. Also, you can combine none and default cases

polosanya (changes requested):
Almost done! Try to implement the task without directly iterating with DOM

src/App.tsx
Pavlo-Petrashevskyi: done

src/App.tsx
```diff
@@ -16,46 +16,120 @@ export const goodsFromServer = [
...
         <button
           type="button"
           className="button is-success is-light"
+          onClick={(e) => {
+            setSortedBy('length');
+            setIsVisible(true);
+            e.currentTarget.classList.remove('is-light');
+
+            return document
+              .querySelector('.is-info')?.classList.add('is-light');
```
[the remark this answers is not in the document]
Pavlo-Petrashevskyi: done

src/App.tsx
```diff
@@ -15,47 +16,108 @@ export const goodsFromServer = [
...
+    sortBy = Sort.NONE,
+    visible = false,
+    reversed = false,
+  ) => {
+    setSortedBy(sortBy);
+    setIsVisible(visible);
+    setIsReversed(reversed);
+  };
+
+  const changeArray = (array: string[]) => {
```
SanyaBratashchuk, line 40:
```suggestion
  const getVisibleGoods = (array: string[]) => {
```

src/App.tsx
```diff
@@ -15,47 +16,108 @@ export const goodsFromServer = [
...
+
+  const sentDataFromButtons = (
+    sortBy = Sort.NONE,
+    visible = false,
+    reversed = false,
+  ) => {
+    setSortedBy(sortBy);
+    setIsVisible(visible);
+    setIsReversed(reversed);
+  };
```
SanyaBratashchuk, line 38: It is better to change state values independently

etojeDenys (approved):
LGTM

src/App.tsx
```diff
@@ -15,47 +16,108 @@ export const goodsFromServer = [
...
+enum Sort {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortedBy, setSortedBy] = useState(Sort.NONE);
+  const [isVisible, setIsVisible] = useState(false);
```
SanyaBratashchuk, line 28: There is no need to hold whether Reset button is visible in state because it depends on values of existing states

SanyaBratashchuk (changes requested):

src/App.tsx
Pavlo-Petrashevskyi: done

src/App.tsx
Pavlo-Petrashevskyi: done

src/App.tsx
```diff
@@ -15,47 +16,104 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={classNames('button', 'is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={() => {
+            setIsReversed(!isReversed);
```
Volodymyr-Melnyk-Fron10der, line 92:
```suggestion
            setIsReversed((prevReversed) => !prevReversed);
```

Volodymyr-Melnyk-Fron10der: it's better to use callback to get a current state value

Volodymyr-Melnyk-Fron10der (approved):

PR #992, opened 2023-06-11
add task solution

NikaNika12:
https://NikaNika12.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,112 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={classNames('button', 'is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={() => setIsReversed(!isReversed)}
```
TarasHoliuk, line 99: if you use current state value it's better to write functional setState:
```suggestion
          onClick={() => setIsReversed((currentIsReversed) => !currentIsReversed)}
```
Also, better to move it to a separate function (in the same way as `reset`)

src/App.tsx
```diff
@@ -15,47 +16,112 @@ export const goodsFromServer = [
...
-        </button>
+        {sortType !== SortType.NONE || isReversed ? (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={reset}
+          >
+            Reset
+          </button>
+        ) : null}
```
TarasHoliuk, line 112: use ternary

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((good1, good2) => {
```
VitaliyBodnar, line 36: good1 and good2 it's not good naming, you can rename them to firstGood and secondGood
or you can use just default name for sort params, like .sort((a, b) => {

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
+    switch (sortType) {
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.NONE:
```
VitaliyBodnar, line 44: don't need to write this empty case, because you have a default case ('return 0')

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
+
+  const reset = () => {
+    setIsReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const reoderedGoods = getReorderedGoods(goodsFromServer, {
+    isReversed,
+    sortType,
+  });
```
VitaliyBodnar, line 69: make sense to use useMemo here!

src/App.tsx
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
```
VitaliyBodnar, line 3: according to the style guide, import styles should be at the bottom of the imports list

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={reset}
+          >
+            Reset
+          </button>
+        ) : (
+          null
+        )}
```
VitaliyBodnar, line 116: don't need to display null
```suggestion
        {sortType !== SortType.NONE || isReversed && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={reset}
          >
            Reset
          </button>
        )}
```

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
       </div>
 
       <ul>
         <ul>
```
VitaliyBodnar, line 120: why do we need the double 'ul' ?

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

src/App.tsx
NikaNika12: мы еще не изучали этот хук. вот я могу его использовать? нужно ли где-то еще его использовать?
const reoderedGoods = useMemo(() => getReorderedGoods(goodsFromServer, {
    isReversed,
    sortType,
  });

src/App.tsx
NikaNika12: подсвечивает красным обьясни как это понять

demetrxx (changes requested):
Hey, looks like you didn't push anything since last review.

If you have any questions - feel free to ask them in fe_chat :)

NikaNika12:
все заменила

src/App.tsx
VitaliyBondarenko1982: also useMemo hook required dependency array.

```
  const reoderedGoods = useMemo(() => getReorderedGoods(goodsFromServer, {
    isReversed,
    sortType,
  }), [isReversed, sortType]);

VitaliyBondarenko1982 (approved):

PR #1001, opened 2023-06-15
rewrite list-of-goods using useState

cephy77:
[DEMO LINK](https://cephy77.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { Products } from './components/Products/Products';
+import { Button } from './components/Button/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -15,49 +17,111 @@ export const goodsFromServer = [
...
+    visibleGoods.sort();
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
naziamoff, line 47: limited number of pre-defined conditions. You sure `if` is the best option here?

src/App.tsx
```diff
@@ -15,49 +17,111 @@ export const goodsFromServer = [
...
+    <div className="App">
       <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
+        <Button
+          styles={{
+            'is-info': true,
+            'is-light': sortType !== SortType.ALPHABET,
+          }}
```
naziamoff, line 78: nope, we don't set styles inline

src/App.tsx
```diff
@@ -15,49 +17,111 @@ export const goodsFromServer = [
...
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const sortGoods = (value: SortType) => {
+    setSortType(sortType === value ? SortType.NONE : value);
```
naziamoff, line 57: not obvious what's happening here

naziamoff (changes requested):
not bad, but some polishing is definitely required

src/App.tsx
```diff
@@ -15,49 +17,111 @@ export const goodsFromServer = [
...
         >
           Sort alphabetically
-        </button>
+        </Button>
 
-        <button
-          type="button"
-          className="button is-success is-light"
+        <Button
+          styles={{
```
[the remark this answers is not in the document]
cephy77: @naziamoff It is used inside the Button component, or should I define className as an argument of the Button tag?

src/App.tsx
```diff
@@ -15,49 +17,116 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const sortABC = () => {
```
dinaminahi, line 69:
```suggestion
  const sortAlphabetically = () => {
```

src/App.tsx
```diff
@@ -15,49 +17,116 @@ export const goodsFromServer = [
...
+  const reverse = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const reset = () => {
+    setIsReversed(false);
+    setSortType(SortType.NONE);
+  };
+
+  const isChanged = isReversed || sortType !== SortType.NONE;
```
dinaminahi, line 86:
```suggestion
  const isOrderChanged = isReversed || sortType !== SortType.NONE;
```

src/App.tsx
```diff
@@ -15,49 +17,116 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+export enum ButtonType {
+  ALPHABET = 'is-info',
+  LENGTH = 'is-success',
+  REVERSE = 'is-warning',
+  RESET = 'is-danger',
```
dinaminahi, line 24: It would be better to name it like:
```suggestion
  INFO = 'is-info',
  SUCCESS = 'is-success',
  WARNING = 'is-warning',
  DANGER = 'is-danger',
```
because these types are related to this component, and for example you want to use another button in another component that would be 'Buy', and you will set type={ButtonType.LENGTH}, it can be strange

dinaminahi (approved):
Well done 👍 
Check few suggestions

PR #1005, opened 2023-06-15
solution

okosohor:
[DEMO LINK](https://okosohor.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,89 @@ export const goodsFromServer = [
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || type !== SortType.NONE) && (
+          <button
+            onClick={() => {
+              changeReverse(false);
+              setType(SortType.NONE);
+            }}
```
naziamoff, line 89: let's create a method for that

naziamoff (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,86 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const [reverse, setReverse] = useState(false);
```
dinaminahi, line 26: It is boolean variable so better name is 
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +16,86 @@ export const goodsFromServer = [
...
+  const [type, setType] = useState(SortType.NONE);
```
dinaminahi, line 27: to make it motre descriptive
```suggestion
  const [sortType, setSortType] = useState(SortType.NONE);
```

src/App.tsx
```diff
@@ -15,47 +16,86 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(reverse || type !== SortType.NONE) && (
```
dinaminahi, line 84: Let's create a variable for this

dinaminahi (changes requested):
few more things fix

33Orange (approved):

PR #1022, opened 2023-06-15
solution

maryanmats:
[DEMO LINK](https://maryanmats.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +16,126 @@ export const goodsFromServer = [
...
+      visibleGoods.sort();
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((itemA, itemB) => itemA.length - itemB.length);
+      break;
+
+    default:
+      break;
+  }
```
33Orange, line 47: You can use 1 sort and change it condition inside callback
```suggestion
  visibleGoods.sort((itemA, itemB) => {
      switch (coniditon) {
        case 1:
          break;
 
        case 2:
          break;

      default:
        break;
    }
  }
```

src/App.tsx
```diff
@@ -15,49 +16,126 @@ export const goodsFromServer = [
...
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setDirection] = useState(false);
+  const [sortType, setSort] = useState(SortType.NONE);
```
33Orange, line 58: Let's follow the naming convention for the useState. 
`const [value, setValue] = useState()`

```suggestion
  const [isReversed, setIsReversed] = useState(false);
  const [sortType, setSortType] = useState(SortType.NONE);
```

src/App.tsx
```diff
@@ -15,49 +16,126 @@ export const goodsFromServer = [
...
+            className={classNames(
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={reverseList}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
```
33Orange, line 115: Let's move this condition to a variable and use it here

33Orange (changes requested):
Almost done

A few moments to fix

PR #1026, opened 2023-06-16
Solved

olehmahera:
[DEMO LINK](https://olehmahera.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,113 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setReverse] = useState(false);
```
33Orange, line 26:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +16,113 @@ export const goodsFromServer = [
...
+          return 0;
+      }
+    });
+
+    if (reversed) {
+      visibleGoods.reverse();
+    }
+
+    return visibleGoods;
+  };
```
33Orange, line 59: Why is this function inside the app component?
It will be created every re-render.
Move it outside `App` component

src/App.tsx
```diff
@@ -15,47 +16,113 @@ export const goodsFromServer = [
...
+
+  const reorderedGoods = getReorderedGoods(
+    goodsFromServer,
+    sortOrder,
+    isReversed,
+  );
+
+  const isDefaultOrder = sortOrder === SortType.NONE
```
33Orange, line 67:
```suggestion
  const isResetButtonActive = sortOrder === SortType.NONE
```

src/App.tsx
```diff
@@ -15,47 +16,113 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(!isDefaultOrder) && (
```
33Orange, line 104: Change the comparison so that you render the component on a positive condition

src/App.tsx
```diff
@@ -15,47 +16,113 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {reorderedGoods
+            .map(good => (
+              <li
+                key={good}
+                data-cy="Good"
+              >
```
33Orange, line 122: Do you need likebreaks when you have only 2 props?
```suggestion
              <li key={good} data-cy="Good">
```

33Orange (changes requested):
Almost done

A few moments to fix

33Orange (approved):
Well done

PR #1030, opened 2023-06-16
solution

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { FC, useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +17,114 @@ export const goodsFromServer = [
...
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      visibleGoods.reverse();
+    }
+
+    return visibleGoods;
+  };
```
33Orange, line 56: Let's remove this function outside your component `app` to avoid recreating every re-render

src/App.tsx
```diff
@@ -15,48 +17,114 @@ export const goodsFromServer = [
...
+
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setReverse] = useState(false);
```
33Orange, line 59:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +17,114 @@ export const goodsFromServer = [
...
+  const shouldShowReset = isReversed || sortType !== SortType.NONE;
+
+  const sortBy = (sortByKey: SortType) => {
+    setSortType(sortByKey);
+  };
```
33Orange, line 64: This function does nothing. You can use `setSortType(SortType.ALPHABET)` without this function

src/App.tsx
```diff
@@ -15,48 +17,114 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getReorderedGoods(
+          goodsFromServer,
+          { sortType, isReversed },
+        ).map((item) => (
```
33Orange, line 123: Let's create a variable for reordered goods
```suggestion
        {reorderedGoods.map((item) => (
```

33Orange (changes requested):
Almost done

A few moments to fix

33Orange (approved):
Good job

PR #1040, opened 2023-06-18
solution

thienlaoo:
[DEMO LINK](https://thienlaoo.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,48 +15,106 @@ export const goodsFromServer = [
...
+  return visibleGoods;
+}
 
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const reverse = () => {
+    setIsReversed(true);
+  };
```
33Orange, line 59: What would be if i click reverse 2 times? List must change to original order. It doesn't work.
Failed test said you same info. Please don't push solutions with failed tests

src/App.tsx
```diff
@@ -14,48 +15,106 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        { (sortType !== SortType.NONE || isReversed) && (
```
33Orange, line 102: Let's add variable for this condition and use it here

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #1051, opened 2023-06-26
all tests good

Mikhaylo-Orlov:
[DEMO LINK](https://Mikhaylo-Orlov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,13 +34,18 @@ export function getReorderedGoods(
 ) {
   const visibleGoods = [...goods];
 
-  if (sortType === SortType.ALPHABET) {
-    visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
-  }
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
 
-  if (sortType === SortType.LENGTH) {
-    visibleGoods.sort((good1, good2) => good1.length - good2.length);
-  }
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
 
   if (isReversed) {
     visibleGoods.reverse();


src/App.tsx
```diff
@@ -15,49 +17,103 @@ export const goodsFromServer = [
...
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== SortType.ALPHABET },
+            )}
+            onClick={() => setSortType(SortType.ALPHABET)}
```
VitaliyBondarenko1982, line 72: it is bad practise to make anonymous arrow function in jsx
consider to make setSort handler which will accept sortType and return arrow function and call here and for sort by length button.
```
  const setSort = (sortType: SortType) => () => setSortType(sortType);
```
```suggestion
            onClick={setSort(SortType.ALPHABET)}
```

src/App.tsx
```diff
@@ -15,49 +17,103 @@ export const goodsFromServer = [
...
-      </ul>
-    </div>
+          <button
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== SortType.LENGTH },
+            )}
+            onClick={() => setSortType(SortType.LENGTH)}
```
VitaliyBondarenko1982, line 84:
```suggestion
            onClick={setSort(SortType.LENGTH)}
```
see comment for sort Sort alphabetically button

src/App.tsx
```diff
@@ -15,49 +17,103 @@ export const goodsFromServer = [
...
+          </button>
+
+          {((sortType !== SortType.NONE) || (isReversed)) && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => {
+                setSortType(SortType.NONE);
+                setIsReversed(false);
+              }}
```
VitaliyBondarenko1982, line 108: same for arrow function in jsx. consider to make handler and pass reference here.
same for reverse button.

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.

src/App.tsx
```diff
@@ -15,49 +17,109 @@ export const goodsFromServer = [
...
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((good1, good2) => good1.length - good2.length);
+  }
```
OleziO, line 43: Replace with switch case construction.

OleziO (changes requested):

PR #1057, opened 2023-07-01
copy solution from previous task and added types

JasterHim:
[DEMO LINK](https://JasterHim.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,121 @@ export const goodsFromServer = [
...
+            : good1.localeCompare(good2);
+
+        default:
+          return 0;
+      }
+    });
+  }
+
+  return preliminaryGoods;
+}
```
VitaliyBodnar, line 59: Your solution looks a little confusing, let's improve it
```suggestion
enum SortType {
  NONE,
  ALPHABET,
  LENGTH,
}
type ReorderOptions = {
  sortType: SortType,
  isReversed: boolean,
};

export function getReorderedGoods(
  goods: string[],
  { sortType, isReversed }: ReorderOptions,
) {
  const visibleGoods = [...goods];

  visibleGoods.sort((good1, good2) => {
    switch (sortType) {
      case (SortType.ALPHABET):
        return good1.localeCompare(good2);

      case (SortType.LENGTH):
        return good1.length - good2.length;

      default:
        return 0;
    }
  });

  if (isReversed) {
    visibleGoods.reverse();
  }

  return visibleGoods;
}
```

VitaliyBodnar (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,118 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== 0 || isReversed) && (
```
OleziO, line 108: Replace with `(sortType !== SortType.None || isReversed)`

OleziO (changes requested):

sTorba24 (approved):
Well done!

PR #1062, opened 2023-07-02
add task solution

irinabolbat:
[DEMO LINK](https://irinabolbat.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+interface Sort {
```
GGsmou, line 19: I think enum will be more suitable instead of interface

*Use enum SortType instead of string literals for possible sorting options (including the default one)*

GGsmou (approved):
GJ, approving!
But I strongly advise checking enum topic of TS, because it's much more suitable for some tasks)

src/App.tsx
anstsot: agree with [comment](https://github.com/mate-academy/react_list-of-goods-hooks/pull/1062/files#r1250579509). try to use `enum`

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
+            'is-warning',
+            { 'is-light': !reversedGoods },
+          )}
+          onClick={() => (
+            !reversedGoods ? (
+              setReversedGoods(true)
+            ) : (
+              setReversedGoods(false)
+            )
+          )}
```
anstsot, line 106:
```suggestion
          onClick={() => setReversedGoods(!reversedGoods)}
```

anstsot (changes requested):
Let’s add a few fixes to improve your code.

anstsot (approved):
Good work, approving:sparkles:

PR #1063, opened 2023-07-03
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { FC, useState } from 'react';
+import className from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +17,112 @@ export const goodsFromServer = [
...
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
```
IvanFesenko, line 48: No need to sort in case SortType is NONE

src/App.tsx
```diff
@@ -15,47 +17,112 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={className('button is-success', {
+            'is-light': sortType !== SortType.LENGTH,
+          })}
+          onClick={() => setSortType(SortType.LENGTH)}
```
IvanFesenko, line 91: It will be better to create separate handler of sortType change instead of anonymous function
```
const onSortTypeChange = (type: SortType) => () => {
  setSortType(type)
}
```
usage:
`onClick={onSortTypeChange(SortType.ALPHABET)}`
and 
`onClick={onSortTypeChange(SortType.LENGTH)}`

src/App.tsx
```diff
@@ -15,47 +17,112 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={className('button is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={() => setIsReversed(!isReversed)}
```
IvanFesenko, line 101: Same here - create separate handler

IvanFesenko (changes requested):
Good job. Lets make some improvements- check my comments

etojeDenys (changes requested):
reverse button does not work

etojeDenys (changes requested):
the reverse button also must be turned off if the user clicks Reset btn

VitaliyBondarenko1982 (approved):

PR #1065, opened 2023-07-03
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn(
+            'button is-info',
+            { 'is-light': sortField !== SortType.Alphabetically },
+          )}
+          onClick={() => setSortField(SortType.Alphabetically)}
```
adv555, line 75: Might be more correct to create a separate function instead of providing an anonymous function on the onClick event

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn(
+            'button is-warning',
+            { 'is-light': !isReversed },
+          )}
+          onClick={() => setIsReversed(!isReversed)}
```
adv555, line 97: I think it could be the same function as for handleSort
```
  const handleSort = (type: typeof SortType[keyof typeof SortType]) => {
    if (type === SortType.SORT_REVERSE) {
      setIsReversed(!isReversed);
    } else {
      setSortField(type);
    }
  };
```

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={cn(
+            'button is-success',
+            { 'is-light': sortField !== SortType.ByLength },
+          )}
+          onClick={() => setSortField(SortType.ByLength)}
```
adv555, line 86: The same

src/App.tsx
```diff
@@ -15,48 +16,109 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortType.Default);
+              setIsReversed(false);
+            }}
```
adv555, line 109: The same

adv555 (approved):
Nice work!✨
some hints on your choice)

qoa11a (approved):

PR #1076, opened 2023-07-04
task done

pos1bl:
https://pos1bl.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import classNames from 'classnames';
+import React, { useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,133 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type Good = string;
+type Sort = SortType | '';
```
DarkMistyRoom, line 21: these types are redundant. You can use string instead of `Good` and it will be more clear to read. And you can add NONE sort type with `''` value to `SortType` enum

src/App.tsx
```diff
@@ -15,48 +17,133 @@ export const goodsFromServer = [
...
+
+enum SortType {
+  SORT_FIELD_ALPHABET = 'alphabet',
+  SORT_FIELD_LENGTH = 'length',
+  SORT_FIELD_REVERSE = 'reverse',
+}
```
DarkMistyRoom, line 27: It isn't the best idea to combine the sort field and reverse values in one enum, as for me. Consider using this enum only for sort fields and creating boolean variable for reverse

DarkMistyRoom (changes requested):
Good job! Check comments

mykhalenych (approved):

PR #1077, opened 2023-07-05
ListSort

Tetiana-Hishchak:
https://Tetiana-Hishchak.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -14,49 +15,124 @@ export const goodsFromServer = [
...
+          return 0;
+      }
+    });
+  }
+
+  return preparedGoods;
+}
 
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
```
etojeDenys, line 66: we can save our enum values here instead of string

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -14,49 +15,126 @@ export const goodsFromServer = [
   'Jam',
   'Garlic',
 ];
+enum SortType {
+  LENGTH = 'length',
+  NAME = 'name',
+  EMTY = '',
```
VitaliyBondarenko1982, line 21:
```suggestion
  EMPTY = 'empty',
```
typo and better to set some value instead empty string.

VitaliyBondarenko1982 (approved):
Well done
One minor comment for future tasks)

PR #1083, opened 2023-07-05
FedenkoListTS

VladFedenko:
https://VladFedenko.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
+import cn from 'classnames';
 import './App.scss';
+import { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,115 @@ export const goodsFromServer = [
...
+    switch (sortField) {
+      case SortType.Alphabeth:
+        return a.localeCompare(b);
+      case SortType.Length:
+        return a.length - b.length;
+      default:
+        return 0;
+    }
+  });
+
```
etojeDenys, line 41: let's avoid sorting when it's not needed
```suggestion
 if(sortField) {
  prepearedGoods.sort((a, b) => {
    switch (sortField) {
      case SortType.Alphabeth:
        return a.localeCompare(b);
      case SortType.Length:
        return a.length - b.length;
      default:
        return 0;
    }
  });
 }
```

src/App.tsx
```diff
@@ -15,48 +16,115 @@ export const goodsFromServer = [
...
+  function makeSetSortField(field: SortType) {
+    return () => setSortField(field);
+  }
+
+  const reset = () => {
+    const isResetButton = sortField !== '' || isReversed;
+
+    return isResetButton;
+  };
```
etojeDenys, line 69: better to just use `sortField !== '' || isReversed` without function

etojeDenys: it's not so complex condition

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+    goodsFromServer, sortField, isReversed,
+  );
+
+  const reverse = () => (
+    setIsReversed(prevIsReversed => !prevIsReversed)
+  );
+
+  function makeSetSortField(field: SortType) {
+    return () => setSortField(field);
+  }
```
etojeDenys, line 65: we can just call setSortField instead of this func

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  Alphabeth = 'alphabet',
+  Length = 'length',
+}
```
etojeDenys, line 22: let's also add some default value for enum

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+
+function getPrepearedGoods(
+  goods: string[],
+  sortField: string,
```
```suggestion
  sortField: SortType,
```

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    prepearedGoods.reverse();
+  }
+
+  return prepearedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
```
etojeDenys, line 52:
```suggestion
  const [sortField, setSortField] = useState<SortType>(SortType.Default)
```

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1089, opened 2023-07-06
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -87,7 +87,7 @@ export const App: React.FC = () => {
         <button
           type="button"
           className={cn('button is-warning', { 'is-light': !reverseType })}
-          onClick={() => setReverseType(reverseType !== true)}
+          onClick={() => setReverseType(prevType => !prevType)}
         >
           Reverse
         </button>


src/App.tsx
```diff
@@ -15,49 +17,92 @@ export const goodsFromServer = [
...
+  if (type[1] === SortType.reverse) {
+    sortedGoods.reverse();
+  }
+
+  return sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType]
+  = React.useState<SortType[]>([SortType.default, SortType.default]);
```
polosanya, line 47:
```suggestion
  const [sortType, setSortType] = useState<SortType[]>(
    [SortType.default, SortType.default]
  );
```
I don't think that it's a good idea to set array of types. Just create a separate state for reverse

polosanya (changes requested):

polosanya (changes requested):
Comments above

src/App.tsx
```diff
@@ -15,49 +17,101 @@ export const goodsFromServer = [
...
+function getSortedGoods(goods: string[], type: SortType, reverse: SortType) {
+  const sortedGoods = [...goods];
+
+  if (type === SortType.byAlphabet) {
+    sortedGoods.sort((good1, good2) => good1.localeCompare(good2));
+  }
+
+  if (type === SortType.byLength) {
+    sortedGoods.sort((good1, good2) => good1.length - good2.length);
+  }
```
OleziO, line 36: Replace with switch case construction

OleziO (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,93 @@ export const goodsFromServer = [
...
+          onClick={() => setSortType(SortType.byLength)}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn('button is-warning', { 'is-light': !reverseType })}
+          onClick={() => setReverseType(reverseType !== true)}
```
Anastasiia-Svintsova, line 90: It's better to use functional setState here to ensure that you are working with the latest state, even if updates are batched or delayed.
```suggestion
          onClick={() => setReverseType(prevType => !prevType)}
```

Anastasiia-Svintsova (approved):

PR #1090, opened 2023-07-07
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,140 @@ export const goodsFromServer = [
...
+      return (isReversed)
+        ? [...visibleGoods].reverse()
+        : [...visibleGoods];
+  }
+}
+
 export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState([...goodsFromServer]);
+  const [activeSort, setActiveSort] = useState(SortType.NONE);
+  const [activeReset, setActiveReset] = useState(false);
```
SanyaBratashchuk, line 55:
```suggestion
```
You don't need this in the state since reset is active when sort is not NONE or isReversed is true. There is no need to hold in the state data which value depends on other state values only

src/App.tsx
```diff
@@ -15,47 +16,140 @@ export const goodsFromServer = [
...
+    setActiveSort(SortType.LENGTH);
+    setActiveReset(true);
+  };
+
+  const handleReverse = () => {
+    setVisibleGoods([...visibleGoods].reverse());
+    setActiveReverse(!activeReverse);
+
+    if (activeSort !== SortType.NONE) {
+      setActiveReset(activeReset);
```
SanyaBratashchuk, line 83: use a functional setState update when new value of state depends on previous

src/App.tsx
```diff
@@ -15,47 +16,140 @@ export const goodsFromServer = [
...
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map(good => (
+            <li data-cy="Good">
```
SanyaBratashchuk, line 149: add keys

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,131 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(activeSort !== SortType.NONE || activeReverse !== false)
```
SanyaBratashchuk, line 122:
```suggestion
        {(activeSort !== SortType.NONE || activeReverse)
```

SanyaBratashchuk (approved):

PR #1091, opened 2023-07-07
react_list-of-goods-hooks

AndriiYelieva:
[DEMO LINK](https://AndriiYelieva.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,10 +17,11 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-const SORT_BY = {
-  ALPHA: 'alpha',
-  LENGTH: 'length',
-};
+enum SortingCase {
+  ALPHA = 'alphabetically',
+  LENGTH = 'length',
+  default = '',
+}
 
 type SortingParams = {
   sortBy: string,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const SORT_BY = {
+  ALPHA: 'alpha',
+  LENGTH: 'length',
+};
```
etojeDenys, line 23: you can use enum instead of simple object

src/App.tsx
```diff
@@ -15,48 +17,112 @@ export const goodsFromServer = [
...
+
+  if (isReverse) {
+    result.reverse();
+  }
+
+  return result;
+}
+
 export const App: React.FC = () => {
+  const [sortBy, setSortBy] = useState('');
```
etojeDenys, line 62: let's add default value for SORT_BY
```suggestion
  const [sortBy, setSortBy] = useState(SORT_BY.default);
```

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #1105, opened 2023-07-10
add task solution

NazarBaraban:
https://nazarbaraban.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,12 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import cn from 'classnames';
+
+enum SortType {
+  Alphabet = 'alphabet',
+  Length = 'length',
+}
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +21,95 @@ export const goodsFromServer = [
...
+  }
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          onClick={() => setSortField(SortType.Alphabet)}
+          className={cn('button', 'is-success', {
```
AlyonaSarapina, line 69:
```suggestion
          className={cn('button', 'is-info', {
```
wrong class name

AlyonaSarapina (changes requested):
Hey! Good job!
There is a small detail to fix🧑‍🔧

AlyonaSarapina (approved):

src/App.tsx
```diff
@@ -15,48 +21,95 @@ export const goodsFromServer = [
...
+
+  if (IsReverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType | undefined>(undefined);
```
YegorKorenkov, line 54: better to use **null** instead of **undefined**

YegorKorenkov (approved):
Good job!

PR #1115, opened 2023-07-10
solution

diana-tuz:
DEMO LINK](https://diana-tuz.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,147 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortFields {
+  byAlfabet = 'alphabet',
+  byLength = 'length',
+  byIndexdesc = 'reverse default',
```
TarasHoliuk, line 22: it shouldn't be here - just reverse array sorted by other properties if it's needed

src/App.tsx
```diff
@@ -15,48 +16,147 @@ export const goodsFromServer = [
...
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortFields.default);
+  const [sortDirection, setSortDirection] = useState(SortDirections.byAsc);
+  const sortGoods = getSortGoods(goodsFromServer, sortField, sortDirection);
+  const sortReverse = () => {
```
TarasHoliuk, line 82: you should only change sortDirection here, don't touch sortField

src/App.tsx
```diff
@@ -15,48 +16,147 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortFields {
+  byAlfabet = 'alphabet',
```
TarasHoliuk, line 20:
```suggestion
enum SortBy {
  Alfabet = 'alphabet',
```
CamelCase for enum keys

src/App.tsx
```diff
@@ -15,48 +16,147 @@ export const goodsFromServer = [
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {sortField !== SortFields.default && (
+          <button
+            onClick={() => {
+              resetSorting();
+            }}
```
TarasHoliuk, line 146:
```suggestion
            onClick={resetSorting}
```

TarasHoliuk (changes requested):

vlad-demchuk (changes requested):
Fix the previous comments discussed in fe_chat

src/App.tsx
```diff
@@ -15,48 +16,136 @@ export const goodsFromServer = [
...
 
+enum SortBy {
+  Alfabet = 'alphabet',
+  Length = 'length',
+  Indexdesc = 'reverse Default',
+  Default = 'Default',
+}
+
+const sortAlfabet = (a: string, b: string, direction: boolean) => {
+  return direction === true
```
Anastasiia-Svintsova, line 27: the direction parameter is already a boolean, you can directly use it in the ternary operator without explicitly comparing it to true

src/App.tsx
```diff
@@ -15,48 +16,136 @@ export const goodsFromServer = [
...
 ];
 
+enum SortBy {
+  Alfabet = 'alphabet',
+  Length = 'length',
+  Indexdesc = 'reverse Default',
+  Default = 'Default',
+}
+
+const sortAlfabet = (a: string, b: string, direction: boolean) => {
```
Anastasiia-Svintsova, line 26: There is a typo here. 
```suggestion
const sortAlphabet = (a: string, b: string, direction: boolean) => {
```

src/App.tsx
```diff
@@ -15,48 +16,136 @@ export const goodsFromServer = [
...
+      }
+    });
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortBy.Default);
+  const [sortDirection, setSortDirection] = useState(true);
```
Anastasiia-Svintsova, line 71: In my opinion, it will be better to use the `isReversed` state for this purpose. The direction can be `asc` or `desc`. But here is a boolean value https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -15,48 +16,136 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortBy {
+  Alfabet = 'alphabet',
+  Length = 'length',
+  Indexdesc = 'reverse Default',
```
Anastasiia-Svintsova, line 22: Why do you need the Indexdesc here?  Remove it, please

src/App.tsx
```diff
@@ -15,48 +16,136 @@ export const goodsFromServer = [
...
-          className="button is-success is-light"
+          className={cn('button is-success', {
+            'is-light': sortField !== SortBy.Length,
+          })}
         >
           Sort by length
         </button>
 
         <button
+          onClick={() => sortReverse()}
```
Anastasiia-Svintsova, line 124: You can simplify it 
```suggestion
          onClick={sortReverse}
```

Anastasiia-Svintsova (changes requested):
Keep going! Please, make sure that you don't use overcomplicated logic. Try to refactor the code to make it easier to read, understand and maintain

src/App.tsx
```diff
@@ -15,48 +16,132 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(true);
+  const sortGoods = getSortGoods(goodsFromServer, sortField, isReversed);
+  const sortReverse = () => {
+    if (isReversed === true) {
+      setIsReversed(false);
+
+      if (sortField === SortBy.Default) {
+        setIsReversed(false);
+      }
+    } else {
```
anastasiiavorobiova, line 82: The else condition 'setIsReversed(true)' is redundant, as it doesn't make any changes.

src/App.tsx
```diff
@@ -15,48 +16,132 @@ export const goodsFromServer = [
...
+  goods: string[],
+  sortField: SortBy,
+  sortDirection: boolean,
+) {
+  const preparedGoods: string[] = [...goods];
+
+  if (sortField) {
+    preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortBy.Alfabet:
```
anastasiiavorobiova, line 51: The Enum key 'Alfabet' is misspelled, it should be 'Alphabet'.

src/App.tsx
```diff
@@ -15,48 +16,132 @@ export const goodsFromServer = [
...
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortBy.Default);
+  const [isReversed, setIsReversed] = useState(true);
+  const sortGoods = getSortGoods(goodsFromServer, sortField, isReversed);
+  const sortReverse = () => {
+    if (isReversed === true) {
```
anastasiiavorobiova, line 76: The condition check 'isReversed === true' is redundant, you can simply check 'if (isReversed)'.

src/App.tsx
```diff
@@ -15,48 +16,132 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortBy {
+  Alfabet = 'alphabet',
```
anastasiiavorobiova, line 20: The Enum key 'Alfabet' is misspelled, it should be 'Alphabet'.

anastasiiavorobiova (changes requested):
Great solution! Added comments to fix

src/App.tsx
```diff
@@ -15,48 +16,128 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortBy.Default);
+  const [isReversed, setIsReversed] = useState(true);
+  const sortGoods = getSortGoods(goodsFromServer, sortField, isReversed);
+  const sortReverse = () => {
+    if (isReversed) {
+      setIsReversed(false);
+    } else {
+      setIsReversed(true);
```
polosanya, line 79:
```suggestion
    setIsReversed(prevValue => !prevValue);
```
It is better to use functional state here with toggling the previous value

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments

PR #1123, opened 2023-07-11
add task solution

NadiaKovalchuk:
[DEMO LINK] https://NadiaKovalchuk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,8 +16,11 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-const SORT_LENGHT = 'length';
-const SORT_ALPHABETICALLY = 'alphabetically';
+enum SortType {
+  ALPH = 'alph',
+  LENGTH = 'length',
+  ALL = '',
+}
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,102 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const SORT_LENGHT = 'length';
+const SORT_ALPHABETICALLY = 'alphabetically';
```
AntonBaranskyi, line 20: Use enum SortType instead of string literals for possible sorting options (including the default one).

So, you have to rewrite it using enums :)

AntonBaranskyi (approved):
Good job, but you should take a look at enums!!

And678 (changes requested):
Requirements: `Use enum SortType instead of string literals for possible sorting options (including the default one)`

anstsot (changes requested):
Please fix previous comments

anstsot (approved):
Good work, approving:sparkles:

PR #1130, opened 2023-07-11
solution

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,48 +27,95 @@ export const goodsFromServer = [
...
+          onClick={() => setSortBy(SortType.SORT_BY_LENGTH)}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn('button', 'is-warning',
+            { 'is-light': isReverse === false })}
```
volodymyr-soltys97, line 95: Instead of comparing 'isReverse' to false, you could use '!isReverse'. This makes the code cleaner and easier to understand.

src/example.jsx
```diff
@@ -0,0 +1,112 @@
+import { useState } from 'react';
```
volodymyr-soltys97, line 1: The code is written in JavaScript but the task requires TypeScript.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

mykhalenych (approved):

PR #1137, opened 2023-07-12
solution

myplekan:
[DEMO LINK](https://myplekan.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,103 @@ export const goodsFromServer = [
...
+interface SortVer {
+  sortField: SortType,
+  reverseField: boolean,
+}
+
+function gerPreparedGoods(
+  goods: string[],
+  { sortField, reverseField }: SortVer,
+) {
+  const preraredGoods = [...goods];
```
volodymyr-soltys97, line 34: There's a typo in the variable name, it should be 'preparedGoods' instead of 'preraredGoods'.

src/App.tsx
```diff
@@ -15,48 +16,103 @@ export const goodsFromServer = [
...
+  length = 'length',
+}
+
+interface SortVer {
+  sortField: SortType,
+  reverseField: boolean,
+}
+
+function gerPreparedGoods(
+  goods: string[],
```
volodymyr-soltys97, line 31: There's a typo in the function name, it should be 'getPreparedGoods' instead of 'gerPreparedGoods'.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

src/App.tsx
```diff
@@ -15,48 +16,103 @@ export const goodsFromServer = [
...
-          className="button is-success is-light"
+          className={cn('button', 'is-success', {
+            'is-light': sortField !== SortType.length,
+          })}
         >
           Sort by length
         </button>
 
         <button
+          onClick={() => (setReverseField(!reverseField))}
```
anastasiiavorobiova, line 91:
```suggestion
          onClick={() => (setReverseField(true))}
```
There is no toggle button, so you can set value directly

anastasiiavorobiova (approved):
Great solution

PR #1141, opened 2023-07-13
Solution

yehor-kor:
[DEMO LINK](https://yehor-kor.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,111 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  alphabet = 'alphabet',
+  length = 'length',
```
polosanya, line 21:
```suggestion
enum SortType {
  Alphabet = 'alphabet',
  Length = 'length',
```
Enum properties should be capitalized
https://www.w3schools.com/typescript/typescript_enums.php

src/App.tsx
```diff
@@ -15,48 +16,111 @@ export const goodsFromServer = [
...
+}
+
+interface SortParams {
+  sortField: string;
```
polosanya, line 25: You can set enum instead of string for better typification

polosanya (approved):
Great!

PR #1147, opened 2023-07-16
Task is done

Konsttantin:
[DEMO LINK](https://Konsttantin.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,126 @@ export const goodsFromServer = [
...
+}
+
+const getVisibleGoods = (
+  goods: typeof goodsFromServer,
+  { option, reversed }: SortOptions,
+) => {
+  let visibleGoods = goods;
+
+  if (option) {
+    visibleGoods = [...visibleGoods].sort((good1, good2) => {
```
VitaliyBodnar, line 38: don't need to overwrite the visibleGoods, because method sort will do that
```suggestion
  let visibleGoods = [...goods];

  if (option) {
    visibleGoods.sort((good1, good2) => {
```

VitaliyBodnar (changes requested):

src/App.tsx
Konsttantin: But it will broke down my conditional rendering in 116 line, what should I do in this case?

src/App.tsx
```diff
@@ -15,48 +17,126 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {visibleGoods !== goodsFromServer
```
VitaliyBodnar, line 116: this condition is wrong because two arrays can't be the same in js. The same can be only links of the arrays.
for example: try to console this expression:
`console.log([] === [])` or 'console.log([ { id: 1 } ] === [ { id: 1 } ])'

VitaliyBodnar (changes requested):
If you have any questions about the comments, please ask them in the Slack chat

src/App.tsx
Konsttantin: Of course I understand it, in my logic I just want to check if visibleGoods variable consists link to original goodsFromServer. Because I return from function either changed copy of original goodsFromServer array or just link to it, that will mean that there aren't any changes comparing to original goods and reset button shouldn't be visible

Konsttantin: I just want to now what approach to conditional rendering in this case will be better if I'll change my function according to your suggestion. Btw, my code also worked.

Konsttantin: I changed my code, but now it seems to me that conditional rendering became less elegant

demetrxx (approved):
Well done 👍 Go ahead!

PR #1151, opened 2023-07-17
Solved

silcki:
https://silcki.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,112 @@ export const goodsFromServer = [
...
+  if (isSorted) {
+    localGoods.reverse();
+  }
+
+  return localGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
+  const [isSorted, setSorted] = useState(false);
```
mykhalenych, line 57: why do you need additional state check if sortField !== SortType.Default

mykhalenych (changes requested):

src/App.tsx
silcki: I need it for the handling reverse
renamed it for better understanding

mykhalenych (approved):

PR #1153, opened 2023-07-18
add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +15,93 @@ export const goodsFromServer = [
...
+    goodsFromServer, sortBy, isReversed,
+  );
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sortBy !== SortType.name ? 'is-light' : ''}`}
```
etojeDenys, line 60: let's use `classNames` package for such complex classes

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,99 @@ export const goodsFromServer = [
...
+        {(sortBy !== SortType.empty || isReversed)
+          && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => {
+                setSortBy(SortType.empty);
+                setIsReversed(false);
+              }}
+            >
```
IvanFesenko, line 102: NOTE: using anonymous function is bad pratice - you should create separate handler and pass to onClick prop reference to it

IvanFesenko (approved):

PR #1162, opened 2023-07-21
solution

vetal-hovenko:
[DEMO LINK](https://vetal-hovenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,13 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import {
+  Good,
+  ReverseParam,
+  SortParam,
+  SortType,
+  SortingParams,
+} from './types';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +22,110 @@ export const goodsFromServer = [
...
+    }
+  };
+
+  const resetGoods = () => {
+    setSortField(SortType.reset);
+    setSortField('');
+    setSortOrder('');
+  };
+
+  const Reset = () => (
```
anastasiiavorobiova, line 77: Inline component 'Reset' is being declared inside another component. This might lead to unnecessary re-rendering and performance issues. Either move 'Reset' outside of 'App' or convert it to a normal function that returns JSX.

src/App.tsx
```diff
@@ -15,48 +22,110 @@ export const goodsFromServer = [
...
+  const reverseGoods = () => {
+    if (sortOrder === SortType.reverse) {
+      setSortOrder('');
+    } else {
+      setSortOrder(SortType.reverse);
+    }
+  };
+
+  const resetGoods = () => {
+    setSortField(SortType.reset);
```
anastasiiavorobiova, line 72: You're invoking setSortField function twice in a row, this can lead to unnecessary rerenders. The second invocation with the empty string argument may not be needed.

src/App.tsx
```diff
@@ -15,48 +22,110 @@ export const goodsFromServer = [
...
+    return newGoods.reverse();
+  }
+
+  return newGoods;
+}
+
+export const App = () => {
+  const [sortField, setSortField] = useState<SortParam>('');
+  const [sortOrder, setSortOrder] = useState<ReverseParam>('');
+  const goodsForSort = preparedGoods(goodsFromServer, {
```
anastasiiavorobiova, line 59: You're running an expensive operation (sorting) directly in the component body. This operation will run on every render, even if the 'sortField' or 'sortOrder' haven't changed. Consider using the useMemo hook to optimize this.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

mykhalenych (approved):

PR #1169, opened 2023-07-26
Add task solution

ArtemBehter:


src/App.tsx
```diff
@@ -15,48 +15,109 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+
+  const sortAlphabetically = () => {
```
darokrk, line 56: Do we need these three functions? sortAlphabetically, reset and sortByLength can be reduced into one function with a parameter e.g.

```
const sortBy = (sortType) => {
  setSortType(sortType);
  setIsReversed(false);
}
```

darokrk (changes requested):
Looks nice 👏🏽  one thing need improvement

src/App.tsx
```diff
@@ -53,18 +55,8 @@ export const App: React.FC = () => {
...
-  };
-
-  const reset = () => {
-    setSortType(SortType.NONE);
-    setIsReversed(false);
-  };
-
-  const sortByLength = () => {
-    setSortType(SortType.LENGTH);
+  const sortBy = (sortType: any) => {
```
darokrk, line 58: instead of `any` you should use your enum types:

`const sortBy = (sortType: SortType.NONE | SortType.ALPHABET | SortType.LENGTH) => {`

darokrk (changes requested):
Looks good, just one more fix to do, and will be perfect 😎

src/App.tsx
```diff
@@ -98,7 +99,9 @@ export const App: React.FC = () => {
         <button
           type="button"
           className={`button is-danger ${isHidden ? 'hidden-reset-button' : 'is-light'}`}
-          onClick={sortBy}
+          onClick={() => {
+            setIsReversed(false); sortBy(SortType.NONE);
```
darokrk, line 103: setIsReversed(false); here is redundant

src/App.tsx
```diff
@@ -55,7 +55,8 @@ export const App: React.FC = () => {
   const [isReversed, setIsReversed] = useState<boolean>(false);
   const [sortType, setSortType] = useState<SortType>(SortType.NONE);
 
-  const sortBy = (sortType: any) => {
+  // eslint-disable-next-line max-len
+  const sortBy = (sortType: SortType.NONE | SortType.ALPHABET | SortType.LENGTH) => {
```
darokrk, line 59: you should use here if,else check is you need it:

```
{
  if (sortType === SortType.NONE) {
   setIsReversed(false);
   setSortType(sortType);
  } else {
   setSortType(sortType);
   setIsReversed(false);
  }
}
```

darokrk (approved):
Give approval 👍🏽  but please fix my last mention before moving next, good job 💪🏽 🔥 !

PR #1178, opened 2023-08-01
add hooks

Misha408:
[DEMO LINK](https://misha408.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,7 +25,7 @@ enum SortType {
 function getPreparedGoods(
   goods: string[],
   sortField: string,
-  isRev: boolean,
+  isReverse: boolean,
 ): string[] {
   const preparedGoods = [...goods];
 
@@ -42,7 +42,7 @@ function getPreparedGoods(
     }
   });
 
-  if (isRev) {
+  if (isReverse) {
     preparedGoods.reverse();
   }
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,89 @@ export const goodsFromServer = [
...
+  if (isRev) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [isRev, setIsReversed] = useState(false);
```
loralevitska, line 54: Don't cut variable naming 
```suggestion
  const [isReverse, setIsReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +17,89 @@ export const goodsFromServer = [
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || isRev) && (
+          <button
+            onClick={() => {
+              setSortField('');
+              setIsReversed(false);
+            }}
```
loralevitska, line 90: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
Almost done) Let's improve

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Well done 🔥

PR #1180, opened 2023-08-03
Solution

Artem193:
https://Artem193.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,25 +16,27 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const rev = 'reverse';
-export const alphabet = 'alphabet';
-export const alphabetRev = 'alphabet reverse';
-export const length = 'length';
-export const lengthRev = 'length reverse';
+export enum Classes {
+  rev = 'reverse',
+  alphabet = 'alphabet',
+  alphabetRev = 'alphabet reverse',
+  length = 'length',
+  lengthRev = 'length reverse',
+}
 
 export const App: React.FC = () => {
   const [visibleGoods, setVisibleGoods] = useState(goodsFromServer);
   const [sortField, setSortField] = useState('');
 
-  const searchClass = (allClass: string) => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,108 @@ export const goodsFromServer = [
...
+
+  const searchClass = (allClass: string) => {
+    if (allClass.includes(rev)) {
+      return true;
+    }
+
+    return false;
+  };
+
+  const classButton = (buttonName: string, classButtonActive: string) => {
```
polosanya, line 37: Same here
image

polosanya (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+export enum Classes {
+  rev = 'reverse',
+  alphabet = 'alphabet',
+  alphabetRev = 'alphabet reverse',
+  length = 'length',
+  lengthRev = 'length reverse',
```
Anastasiia-Svintsova, line 24: You don't need separate values for reversed values. Refactor the code to avoid it 
```suggestion
  length = 'length',
```

src/App.tsx
```diff
@@ -15,48 +16,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+export enum Classes {
```
Anastasiia-Svintsova, line 19: The `sortBy` name will be more appropriate here 
```suggestion
export enum sortBy {
```

src/App.tsx
```diff
@@ -15,48 +16,112 @@ export const goodsFromServer = [
...
+  rev = 'reverse',
```
Anastasiia-Svintsova, line 20: You don't need the `rev` here. Add an additional state `isReversed` inside the App component

Anastasiia-Svintsova (changes requested):
Hey! The logic looks too complicated now. Try to refactor the code according to my comments

PR #1185, opened 2023-08-06
added list goods

Ilya-Osoben:
[DEMO LINK](https://Ilya-Osoben.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
 ];
 
+type Goods = string;
+type Reverse = boolean;
+
+enum Sortfield {
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+  DEFAULT = '',
+}
```
VitaliyBodnar, line 26: enum's items shoud be in CamelCase 
```suggestion
enum Sortfield {
  Alphabet = 'alphabet',
  Length = 'length',
  Default = '',
}
```

src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
+);
+
+function getPrepareGoods(
+  goods: Goods[],
+  sortGood: Sortfield | undefined,
+  reverse: Reverse,
+): Goods[] {
+  const prepareGoods: Goods[] = [...goods];
+
+  if (prepareGoods) {
```
VitaliyBodnar, line 49: this condition will always return true because you already applied `prepareGoods` as [...goods]
arrays is always be true: [] === true or [{}] === true
in this case, you can check a prepareGoods.length

src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
+    prepareGoods.reverse();
+  }
+
+  return prepareGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortGood, setSortGood] = useState<Sortfield>(Sortfield.DEFAULT);
+  const [reverse, setReverse] = useState<Reverse>(false);
+  const visibleGoods = getPrepareGoods(goodsFromServer, sortGood, reverse);
```
VitaliyBodnar, line 74: make sense to use useMemo here, because it's essential to performance

src/App.tsx
```diff
@@ -15,49 +16,108 @@ export const goodsFromServer = [
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortGood || reverse) && (
+          <button
+            onClick={() => {
+              setSortGood(Sortfield.DEFAULT);
+              setReverse(false);
```
VitaliyBodnar, line 110: it is bad practise to make anonymous arrow functions in jsx.

VitaliyBodnar (changes requested):
Almost done!

src/App.tsx
```diff
@@ -15,49 +16,110 @@ export const goodsFromServer = [
...
+    ))}
+  </ul>
+);
+
+function getPrepareGoods(
+  goods: Goods[],
+  sortGood: Sortfield | undefined,
+  reverse: Reverse,
+): Goods[] {
+  const prepareGoods: Goods[] = [...goods];
```
sTorba24, line 47:
```suggestion
  const preparedGoods: Goods[] = [...goods];
```

src/App.tsx
```diff
@@ -15,49 +16,110 @@ export const goodsFromServer = [
...
+
+  return prepareGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortGood, setSortGood] = useState<Sortfield>(Sortfield.Default);
+  const [reverse, setReverse] = useState<Reverse>(false);
+  const visibleGoods = getPrepareGoods(goodsFromServer, sortGood, reverse);
+
+  const getReset = (): void => {
```
sTorba24, line 76:
```suggestion
  const handleReset = (): void => {
```

sTorba24: https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

sTorba24 (changes requested):
Almost good! Lets make some improvements

src/App.tsx
Ilya-Osoben: Thank you this is really helpful

src/App.tsx
```diff
@@ -15,49 +16,110 @@ export const goodsFromServer = [
...
+);
+
+function getPrepareGoods(
+  goods: Goods[],
+  sortGood: SortType | undefined,
+  reverse: Reverse,
+): Goods[] {
+  const preparedGoods: Goods[] = [...goods];
+
+  if (preparedGoods.length !== 0) {
```
sTorba24, line 49:
```suggestion
  if (preparedGoods.length) {
```

src/App.tsx
sTorba24: image

still not resolved

sTorba24 (approved):

PR #1199, opened 2023-08-11
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,129 @@ export const goodsFromServer = [
...
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortBy.Default);
+  const [isReversed, setIsReversed] = useState(true);
+  const sortGoods = getSortGoods(goodsFromServer, sortField, isReversed);
+  const sortReverse = () => {
+    if (isReversed) {
```
volodymyr-soltys97, line 76: Instead of using an if-else statement to switch the value of 'isReversed', you can simply set it to the opposite of its current value using '!'.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1200, opened 2023-08-15
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +15,114 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type IsReserved = boolean;
```
darokrk, line 24: The type alias 'IsReserved' is not necessary. You can use 'boolean' directly in the function signature for 'getReorderedGoods'.

src/App.tsx
```diff
@@ -15,47 +15,114 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setReserved] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE as SortType);
+
+  const getReorderedGoods = (
```
darokrk, line 30: you should move this function top, outside of App component to avoid re rendering every time

src/App.tsx
```diff
@@ -15,47 +15,114 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
darokrk, line 109: you can store this check statement into variable before return () as e.g.:

`const showResetButton = isReversed || sortType !== SortType.NONE;`

and then use it here to make it cleaner

src/App.tsx
```diff
@@ -15,47 +15,114 @@ export const goodsFromServer = [
...
+    });
+
+    if (oldIsReversed) {
+      visibleGoods.reverse();
+    }
+
+    return visibleGoods;
+  };
+
+  const sortAlphabet = () => {
```
darokrk, line 55: you can combine sortAlphabet and sortLength into one function and pass sort parameter, e.g.

```
const sortBy = (sortType) => {
  setSortType(sortType)
}
```

then you will have one function instead of two :)

darokrk (changes requested):
Looks good 👍🏽  few mentions left 🛠️

darokrk (approved):
Good job 🔥 👏 !

PR #1221, opened 2023-08-26
list of goods ts

Orchibald:
[DEMO LINK](https://orchibald.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,48 +16,99 @@ export const goodsFromServer = [
...
+  if (toReverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortType.SORT_FIELD_NONE);
+  const [reversedField, setReversedField] = useState(false);
```
lerastarynets, line 56: It is better to use more descriptive names for your state variables. For example, 'isReversed' instead of 'reversedField'.

src/App.tsx
```diff
@@ -15,48 +16,99 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField !== '' || reversedField !== false) && (
```
lerastarynets, line 92: pls see [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
        {(sortField || reversedField) && (
```

src/App.tsx
```diff
@@ -15,48 +16,99 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortType.SORT_FIELD_NONE);
+              setReversedField(false);
+            }}
```
lerastarynets, line 99: Pls move the function inside `onClick` handler to a separate const and use as a callback here

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

lerastarynets (approved):
You're on the right track!

PR #1224, opened 2023-08-29
solution

artem-shuhailo:
[DEMO LINK](https://artem-shuhailo.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +15,110 @@ export const goodsFromServer = [
...
+  if (reverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortField] = useState(SortType.NONE);
+  const [reverse, setSortReverse] = useState(false);
```
loralevitska, line 59: Rename all boolean variables using [tips](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)
```suggestion
  const [isReverse, setIsReverse] = useState(false);
```

src/App.tsx
```diff
@@ -15,48 +15,110 @@ export const goodsFromServer = [
...
+    setSortReverse(false);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sortType !== SortType.ALPHABET && 'is-light'}`}
```
loralevitska, line 74: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally. Fix all cases

src/App.tsx
```diff
@@ -15,48 +15,110 @@ export const goodsFromServer = [
...
+  if (sortType) {
+    preparedGoods.sort((good1, good2) => {
+      switch (sortType) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+
+        default:
```
Viktor-Kost, line 44: in a sorting function, you should return 0 if the two elements being compared are equal, positive value if the first is greater, and negative value if the second is greater.

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

src/App.tsx
```diff
@@ -15,48 +15,110 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`button is-warning ${!reverse && 'is-light'}`}
+          onClick={() => {
+            setSortReverse(!reverse);
+          }}
```
Viktor-Kost, line 97: 1. It's better to create own handler functions for onClick
2. It's better not use getter inside setter at useState.
3. Do the similar actions to handle sort clicks (above)
```
const handleReverseClick = () => {
  setSortReverse(prev => !prev);
}
```

```suggestion
          onClick={handleReverseClick}
```

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

PR #1234, opened 2023-08-31
add solution

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,34 +25,35 @@ export const App: React.FC = () => {
   const sortBy = (value: string) => {
     if (value === SORT_BY.alphabetically) {
       if (isReversed) {
-        setGoodsArr(prevState => [...prevState].sort((a, b) => b.localeCompare(a)));
+        [...goodsFromServer].sort((a, b) => b.localeCompare(a));
       } else {
-        setGoodsArr(prevState => [...prevState].sort((a, b) => a.localeCompare(b)));
+        [...goodsFromServer].sort((a, b) => a.localeCompare(b));
       }
 
       setSortValue(SORT_BY.alphabetically);
-      setResetValue(true);
 
       return;
     }
 
-    if (isReversed) {
-      setGoodsArr(prevState => [...prevState].sort((a, b) => b.length - a.length));
-    } else {
-      setGoodsArr(prevState => [...prevState].sort((a, b) => a.length - b.length));
+    if (value === SORT_BY.length) {
+      if (isReversed) {
+        [...goodsFromServer].sort((a, b) => b.length - a.length);
+      } else {
+        [...goodsFromServer].sort((a, b) => a.length - b.length);
+      }
+
+      setSortValue(SORT_BY.length);
+
+      return;
     }
 
-    setSortValue(SORT_BY.length);
-    setResetValue(true);
-  };
-
-  const reset = () => {
-    setGoodsArr(goodsFromServer);
-    setIsReversed(false);
     setSortValue(SORT_BY.none);
+    setIsReversed(false);
   };
 
   const reverse = () => setIsReversed(prevIsReversed => !prevIsReversed);
+
+  const reset = () => sortBy(SORT_BY.none);
 
   return (
     <div className="section content">
@@ -87,7 +88,7 @@ export const App: React.FC = () => {
           Reverse
         </button>
 
-        {resetValue && (
+        {!isReversed && (
           <button
             type="button"
             className="button is-danger is-light"
@@ -99,7 +100,7 @@ export const App: React.FC = () => {
       </div>
 
       <ul>
-        {goodsArr.map(good => (
+        {goodsFromServer.map(good => (
           <li data-cy="Good" key={good}>
             {good}
           </li>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,7 @@ import React, { useState } from 'react';
 import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -16,47 +18,108 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [goodsArr, setGoodsArr] = useState(goodsFromServer);
+  const [sortValue, setSortValue] = useState('');
+  const [reverseValue, setReverseValue] = useState(false);
+  const [resetValue, setResetValue] = useState(false);
+
+  const SORT_BY_ALPHABETICALLY = 'alphabetically';
+  const SORT_BY_LENGTH = 'length';
```
maxim2310, line 27: nove it outside the component

src/App.tsx
```diff
@@ -16,47 +18,108 @@ export const goodsFromServer = [
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {resetValue && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => sortBy('')}
```
maxim2310, line 110: this is have no logic for reset state you call func sortBy, please create separate func for this logic and remove this logic from sortBy

maxim2310 (changes requested):

src/variables/sortBy.ts
```diff
@@ -0,0 +1,4 @@
+export const variables = {
+  sort_by_alphabetically: 'alphabetically',
+  sort_by_length: 'length',
+};
```
VitaliyBondarenko1982, line 4:
```suggestion
export const SORT_BY = {
  alphabetically: 'alphabetically',
  length: 'length',
  none: 'none',
};
```
suggest improve naming. and make value for initial state instead empty string.

src/App.tsx
```diff
@@ -16,47 +19,103 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [goodsArr, setGoodsArr] = useState(goodsFromServer);
+  const [sortValue, setSortValue] = useState('');
+  const [reverseValue, setReverseValue] = useState(false);
+  const [resetValue, setResetValue] = useState(false);
```
VitaliyBondarenko1982, line 25:
```suggestion
  const [sortValue, setSortValue] = useState(SORT_BY.none);
  const [isReversed, setIsReversed] = useState(false);
```
redundant to keep is state resetValue and goods.

change just sortValue and isReversed and based on it make function outside of component whid will accept goodsFromServer, sortValue, isReversed states and return sortedGoods.

src/App.tsx
```diff
@@ -16,47 +19,103 @@ export const goodsFromServer = [
...
+  const reverse = () => {
+    if (!sortValue) {
+      setResetValue(!resetValue);
+    } else {
+      setResetValue(true);
+    }
+
+    setGoodsArr(prevState => [...prevState].reverse());
+    setReverseValue(!reverseValue);
+  };
```
VitaliyBondarenko1982, line 67:
```suggestion
  const reverse = () => setIsReversed(prevIsReversed => !prevIsReversed);
```
click on reverse button should only change isReversed state on opposite

src/App.tsx
```diff
@@ -16,47 +19,103 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={cn('button is-success', {
+            'is-light': sortValue !== variables.sort_by_length,
+          })}
+          onClick={() => sortBy(variables.sort_by_length)}
```
VitaliyBondarenko1982, line 87:
```suggestion
          onClick={() => sortSortValue(SORT_BY.length)}
```
click on sort button should only set proper sort value in state.

VitaliyBondarenko1982 (changes requested):
Please check comments and refactor your solution.

src/App.tsx
dimarogkov: Вийшла якась каша, дайте більш зрозуміліший коментар

etojeDenys (approved):
.

src/variables/sortBy.ts
```diff
@@ -0,0 +1,5 @@
+export const SORT_BY = {
+  alphabetically: 'alphabetically',
+  length: 'length',
+  none: 'none',
+};
```
etojeDenys, line 5: let's use an enum here for sorting type instead of the object

etojeDenys (changes requested):
feel free to ask in the chat if you have some questions
nit:

src/App.tsx
VitaliyBodnar: if something doesn't work out, then please ask you questions in a fe_pt chat

src/App.tsx
```diff
@@ -16,47 +19,92 @@ export const goodsFromServer = [
...
+      if (isReversed) {
+        [...goodsFromServer].sort((a, b) => b.length - a.length);
+      } else {
+        [...goodsFromServer].sort((a, b) => a.length - b.length);
+      }
+
+      setSortValue(SortBY.length);
+
+      return;
+    }
```
VitaliyBodnar, line 48: let's improve you solution 
- make sense to use switch case statement here instead of many `if` blocks
- you need to apply reveres after sorting, for instance 

`goodsFromServer.sort()`
`if (isReversed) {`
`  goodsFromServer.reverse()`
`}`

VitaliyBodnar (changes requested):

PR #1235, opened 2023-08-31
solution

IhorVerba:
[DEMO LINK](https://IhorVerba.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import cn from 'classnames';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { SortField } from './Types/SortField';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,48 +17,101 @@ export const goodsFromServer = [
...
+    preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortField.SORT_FIELD_LENGTH:
+          return good1.length - good2.length;
+
+        case SortField.SORT_FIELD_ALPHABET:
+          return good1.localeCompare(good2);
+
+        default:
+          return 0;
```
maxim2310, line 37:
```suggestion
          return goods;
```

src/App.tsx
```diff
@@ -15,48 +17,101 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField || reverseField) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortField.EMPTY);
+              setReverseField(false);
+            }}
```
maxim2310, line 99: create func reset for this

maxim2310 (changes requested):

src/App.tsx
IhorVerba: > 

It is allowed in Ts to use 0 to compare strings when you need to return an unchanged array, isn't it?

anastasiiavorobiova (approved):
Awesome!

PR #1238, opened 2023-09-02
Refactored react_list-of-goods from js to ts

davinnchii:
[DEMO LINK](https://davinnchii.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,89 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  length = 'length',
+  alphabet = 'alphabet',
+  noSort = '',
+}
```
alexander-ignatow, line 23: Good, but please capitalize enum values (`length` -> `Length`, etc)

src/App.tsx
```diff
@@ -15,48 +16,89 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goods.map((good, _, arr) => (
+          <li
+            data-cy="Good"
+            key={`${good} from ${arr}`}
```
alexander-ignatow, line 97: Having just `good` for the key is okay here
In any case the part "from ${arr}" will not be unique for the goods, so you can omit it and make all mapping simpler.

src/App.tsx
```diff
@@ -15,48 +16,89 @@ export const goodsFromServer = [
...
           Reverse
         </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField !== '' || isReversed) && (
```
alexander-ignatow, line 79: An empty string is a `falsy` value - and you use it directly in your sort method (ln 28)
Please be consistent with this

alexander-ignatow (changes requested):
Good 👍 - please fix some remaining issues

Yevheniia-Sid (approved):
Good job!👍

dffUqp (approved):
GJ✔️

PR #1240, opened 2023-09-02
Solution

andrijdudar:
[DEMO LINK](https://andrijdudar.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import GoodList from './components/goodList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,48 +17,100 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+const SORT_FIELD_ALPHABETICALLY = 'name';
+const SORT_FIELD_LENGTH = 'length';
```
etojeDenys, line 21: let's move sort values to some enum instead of the vars

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +17,102 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+const SORT_FIELD = {
+  ALPHABETICALLY: 'name',
+  LENGTH: 'length',
+};
```
etojeDenys, line 23: let's use enum instead of the object

etojeDenys (approved):
lgtm

PR #1243, opened 2023-09-04
add solution

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,48 +16,108 @@ export const goodsFromServer = [
...
+          throw new Error('Unbelievable error');
+      }
+    });
+  }
+
+  if (isReverse) {
+    udatedGoods.reverse();
+  }
+
+  return udatedGoods;
```
IvanFesenko, line 56: Can omit a bit
```suggestion
  const udatedGoods = [...goods];

  if (sortType) {
    udatedGoods.sort((good1, good2) => {
      switch (sortType) {
        case SortType.leng:
          return good1.length - good2.length;
        case SortType.alph:
          return good1.localeCompare(good2);
        default:
          throw new Error('Unbelievable error');
      }
    });
  }

  if (isReverse) {
    udatedGoods.reverse();
  }

  return udatedGoods;
```

IvanFesenko (approved):
well done

PR #1248, opened 2023-09-04
reimplement list_of_goods app using TS

kaavka:


ivannaromanovych (approved):
Good job!!

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  SortByAlphabet = 'alphabet',
+  SortByLength = 'length',
+  DefaultValue = '',
+}
```
roman-mirzoian, line 23: In the name of the `enam`, you already indicated that it will be sorting, so the names can be made a little shorter:
```suggestion
enum SortType {
  ByAlphabet = 'alphabet',
  ByLength = 'length',
  DefaultValue = '',
}
```

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+
+interface SortingOptions {
+  sortType: SortType,
+  isReversed: boolean,
+}
+
+function getSortedGoodsList(
+  goods: string[],
+  { sortType, isReversed }: SortingOptions,
```
roman-mirzoian, line 32: Why do you pass the second and third arguments as one object?
What is the point in this?

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+  return sortedGoodList;
+}
+
 export const App: React.FC = () => {
+  const [sortingOptions, setSortingOptions] = useState<SortingOptions>({
+    sortType: SortType.DefaultValue,
+    isReversed: false,
+  });
+
+  const visibleGoodsList = getSortedGoodsList(
```
roman-mirzoian, line 66:
```suggestion
  const sortedGoods = getSortedGoodsList(
```

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+  };
+
+  const handlerSetIsReversed = () => {
+    setSortingOptions((prevState: SortingOptions) => ({
+      ...prevState,
+      isReversed: !prevState.isReversed,
+    }));
+  };
+
+  const resetSortingOptions = () => {
```
roman-mirzoian, line 85:
```suggestion
  const handleButtonReset = () => {
```

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+  );
+
+  const handlerSetSortType = (sortType: SortType) => () => {
+    setSortingOptions((prevState: SortingOptions) => ({
+      ...prevState,
+      sortType,
+    }));
+  };
+
+  const handlerSetIsReversed = () => {
```
roman-mirzoian, line 78:
```suggestion
  const handleButtonReverse = () => {
```

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+    sortType: SortType.DefaultValue,
+    isReversed: false,
+  });
+
+  const visibleGoodsList = getSortedGoodsList(
+    goodsFromServer,
+    sortingOptions,
+  );
+
+  const handlerSetSortType = (sortType: SortType) => () => {
```
roman-mirzoian, line 71: https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364
```suggestion
  const handleSortTypeChange = (sortType: SortType) => () => {
```

roman-mirzoian (changes requested):
There is also no link to the demo in the PR description.

src/App.tsx
kaavka: Today we had a conversation with Anna about this. In case when we are destructuring them if we would have situation when there would be much more sorting options, we wouldn't care about order. If we provide them as arguments, we should be confident that we receive them in the same order that we provided.
@roman-mirzoian

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+  };
+
+  const handleButtonReset = () => {
+    setSortingOptions({
+      sortType: SortType.DefaultValue,
+      isReversed: false,
+    });
+  };
+
+  const isSortingOptionsInstalled = (
```
Moroz-Dmytro, line 92:
```suggestion
  const isResetButtonVisible = (
```

src/App.tsx
```diff
@@ -15,48 +16,140 @@ export const goodsFromServer = [
...
+        {isSortingOptionsInstalled
+          && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={handleButtonReset}
+            >
+              Reset
+            </button>
+          )}
```
Moroz-Dmytro, line 144:
```suggestion
        {isSortingOptionsInstalled && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={handleButtonReset}
          >
            Reset
          </button>
        )}
```

Moroz-Dmytro (approved):

PR #1264, opened 2023-09-05
add sort functionality to list of goods

Yevheniia-Sid:
[DEMO LINK](https://Yevheniia-Sid.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer : string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


Taras2705 (approved):
good job!!!

src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
...
+}
+
+interface SortFields {
+  sortField: SortType,
+  isReverse: boolean,
+}
+
+function getPreparedGoods(
+  goods: string[],
+  { sortField, isReverse } :SortFields,
```
dffUqp, line 32:
```suggestion
  { sortField, isReverse }: SortFields,
```

src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
...
+    preparedGoods.sort((goodA, goodB) => {
+      switch (sortField) {
+        case SortType.Alphabet:
+          return goodA.localeCompare(goodB);
+
+        case SortType.Length:
+          return goodA.length - goodB.length;
+
+        default:
+          return 0;
```
dffUqp, line 46: In that case sort method would be called even if it's not needed.
Consider moving `switch...case` outside of the sort callback

src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
...
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
+  const [isReverse, setIsReverse] = useState<boolean>(false);
+  const visibleGoods: string[] = getPreparedGoods(goodsFromServer,
+    { sortField, isReverse });
```
dffUqp, line 62: Why are you putting the last two arguments into the object?

If arguments order bother you, then it's better to put everything in an object.

src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField !== SortType.Default || isReverse) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortType.Default);
+              setIsReverse(false);
+            }}
```
dffUqp, line 104: better to create a separate handler for this

dffUqp (changes requested):
Almost done:smiley:

src/App.tsx
Yevheniia-Sid: > Why are you putting the last two arguments into the object?
> 
> If arguments order bother you, then it's better to put everything in an object.


Thanks for the advice. But I don’t completely agree with you, because I think that the array is always there, and the sorting parameters can change and expand, and it’s better to pass them as an object.

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
+  const [isReverse, setIsReverse] = useState<boolean>(false);
+  const visibleGoods: string[] = getPreparedGoods(goodsFromServer,
+    { sortField, isReverse });
```
BudnikOleksii, line 59:
```suggestion
  const visibleGoods: string[] = getPreparedGoods(
    goodsFromServer,
    { sortField, isReverse },
  );
```

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
+
+  const handlerOnReset = () => {
```
```suggestion
  const handleReset = () => {
```
Prefix `handle` for a handler and the prefix `on` if it's a prop. You don't need to combine them.
You can read about convention [here](https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364)

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn('button', 'is-info', {
+            'is-light': sortField !== SortType.Alphabet,
+          })}
```
BudnikOleksii, line 73:
```suggestion
          className={cn(
            'button',
            'is-info',
            { 'is-light': sortField !== SortType.Alphabet },
          )}
```
Let's improve the code style a bit, change in all places

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField !== SortType.Default || isReverse) && (
```
BudnikOleksii, line 99: Let's move `sortField !== SortType.Default || isReverse` to a variable so you don't need to combine logical operators

BudnikOleksii (changes requested):
Good work 👍 
Only a few minor fixes, check the comments

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
+
+  if (isReverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
```
roman-mirzoian, line 56:
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.Default);
```

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
+  const [isReverse, setIsReverse] = useState<boolean>(false);
```
roman-mirzoian, line 57:
```suggestion
  const [isReversed, setIsReversed] = useState<boolean>(false);
```

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
+  const visibleGoods: string[] = getPreparedGoods(
```
roman-mirzoian, line 58:
```suggestion
  const sortedGoods: string[] = getPreparedGoods(
```

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
+    goodsFromServer,
+    { sortField, isReverse },
+  );
+
+  const isShowReset = sortField !== SortType.Default || isReverse;
+
+  const handleReset = () => {
```
roman-mirzoian, line 65:
```suggestion
  const handleButtonReset = () => {
```

src/App.tsx
```diff
@@ -15,48 +16,116 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map(good => (
+          <li
+            data-cy="Good"
+            key={good}
+          >
```
roman-mirzoian, line 125:
```suggestion
          <li data-cy="Good" key={good}>
```

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #1290, opened 2023-09-06
React List Hooks

olegpaunov:
https://olegpaunov.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
 ];
 
+type Goods = string[];
+
+const SORT_FIELD_BY_LETTERS = 'letters';
+const SORT_FILED_BY_LENGTH = 'length';
+
+function getPreparedGoods(
+  goods: Goods,
+  { sortField, reverseField }: { sortField: string, reverseField: boolean },
```
roman-mirzoian, line 26: Why are you passing the second and third arguments as one object?

roman-mirzoian, line 26:
```suggestion
  { sortType, isReversed }: { sortType: string, isReversed: boolean },
```

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
+
+  if (reverseField) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
```
roman-mirzoian, line 51:
```suggestion
  const [sortType, setSortType] = useState(SortTypes.Default);
```

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type Goods = string[];
+
+const SORT_FIELD_BY_LETTERS = 'letters';
+const SORT_FILED_BY_LENGTH = 'length';
```
roman-mirzoian, line 22: Since we're already using TypeScript here, it's worth storing it via `enam`, adding a default value (empty string) as well.

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reverseField, setReverseField] = useState(false);
+  const visibleGoods = getPreparedGoods(
```
roman-mirzoian, line 53: In any case, you will see the entire list, so the name is incorrect
```suggestion
  const sortedGoods = getPreparedGoods(
```

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
+          onClick={() => setSortField(SORT_FIELD_BY_LETTERS)}
           type="button"
-          className="button is-info is-light"
+          className={cn('button is-info', {
+            'is-light': sortField !== SORT_FIELD_BY_LETTERS,
+          })}
```
roman-mirzoian, line 66: If we use the classnames library, we should use it by dividing classes into different parts if possible.
```suggestion
          className={cn('button', 'is-info', {
            'is-light': sortField !== SORT_FIELD_BY_LETTERS,
          })}
```
Fix this in all code.

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type Goods = string[];
+
+const SORT_FIELD_BY_LETTERS = 'letters';
```
roman-mirzoian, line 21: The name of the button for this type is called "Sort alphabetically", so it is better to stick to the same style in the name.

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || reverseField) && (
```
roman-mirzoian, line 91: This condition can be transferred to a separate variable for better readability.

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
+  if (reverseField) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reverseField, setReverseField] = useState(false);
```
roman-mirzoian, line 52:
```suggestion
  const [isListReversed, setIsListReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,47 +16,97 @@ export const goodsFromServer = [
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || reverseField) && (
+          <button
+            onClick={() => {
+              setSortField('');
+              setReverseField(false);
+            }}
```
roman-mirzoian, line 96: This callback can be moved to a separate handler for better readability.

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -15,47 +16,106 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type Goods = string[];
+
+enum Sorts {
+  sortAlphabetically = 'letters',
+  sortFieldByLength = 'length',
+}
```
roman-mirzoian, line 24:
```suggestion
enum SortTypes {
  alphabetically = 'alphabetically',
  byLength = 'length',
}
```

src/App.tsx
```diff
@@ -15,47 +16,106 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState('');
+  const [isListReversed, setIsListReversed] = useState(false);
+  const sortedGoods = getPreparedGoods(
+    goodsFromServer,
+    sortType,
+    isListReversed,
+  );
+  const CONDITION_FIELD = sortType || isListReversed;
```
roman-mirzoian, line 61:
```suggestion
  const isResetButtonActive = sortType || isListReversed;
```

src/App.tsx
```diff
@@ -15,47 +16,106 @@ export const goodsFromServer = [
...
+
+  function handlerResetClick() {
```
roman-mirzoian, line 63:
```suggestion
  function handleButtonReset() {
```

src/App.tsx
```diff
@@ -15,47 +16,106 @@ export const goodsFromServer = [
...
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {CONDITION_FIELD && (
+          <button
+            onClick={() => handlerResetClick()}
```
roman-mirzoian, line 103:
```suggestion
            onClick={handleButtonReset}
```

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #1291, opened 2023-09-06
task solution

HubertKlamann:
[DEMO LINK](https://HubertKlamann.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


klaudia-klis (approved):
Looks good to me 👍🏼

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, handleSort] = useState(SortType.NONE);
```
JoyCoffeeAddict, line 55: Typically we use a convention to name `useState` variables as: `[sortType, setSortType]` see [here](https://mate-academy.github.io/fe-program/js/extra/react-typescript#usestate) 
we use a prefix `handle` for custom methods.

JoyCoffeeAddict (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {reorderedGoods.map(item => {
+          return (
+            <li data-cy="Good">{item}</li>
```
pawel-peksa, line 116: When we render lists in React we have to remember about adding unique keys to each item. For more info please refer to:
https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn({
+            'button is-warning': true,
+            'is-light': isReversed !== true,
+          })}
+          onClick={() => setIsReversed(!isReversed)}
```
pawel-peksa, line 96:
```suggestion
onClick={() => setReverseField(isReversed => !isReversed)}
```

pawel-peksa, line 96:
```suggestion
          onClick={() => setReverseField(isReversed => !isReversed)}
```

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed)
```
pawel-peksa, line 101: Nitpick: Sometimes when `render` becomes overcrowded or the logical assesment is complicated we can move it to it's separate variable. For example:
before render you define the following:
```typescript
const isResetButtonVisible = sortType !== SortType.NONE || isReversed
```
and then in `render` you have:
```typescript
isResetButtonVisible && (
            <button
              type="button"
              onClick={() => handleReset()}
            >
              Reset
            </button>
          )}
```

It definitely helps with readability, although it is not strictly required here. 
Just letting you know about different approaches :)

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
       <ul>
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {reorderedGoods.map(item => {
```
pawel-peksa, line 114: Suggestion: If the body of fat arrow function consists of only one line you can easily omit curly braces and return statement

src/App.tsx
```diff
@@ -15,48 +16,106 @@ export const goodsFromServer = [
...
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goodsFromServer];
```
pawel-peksa, line 33: I like the fact that you are aware of mutating behavior of the Array methods you are using here and because of that you have created a copy of data.

"fun" fact => Recently we have `toSorted` metod available in JS which actually creates a shallow copy of array and returns sorted data without mutation of the array it was used on.

pawel-peksa (changes requested):
Overall code looks good. 
Good job 👍 
One thing to keep in mind is to use keys when rendering lists in React. (performance)
Other comments are really minor or just small suggestions.

PR #1292, opened 2023-09-06
sol

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +15,131 @@ export const goodsFromServer = [
...
+      isReversed: false,
+    });
+  };
+
+  const sortByLength = () => {
+    setReorderOptions({
+      sortType: SortType.LENGTH,
+      isReversed: false,
+    });
+  };
```
domalewski, line 80: I have a rule that if something is repeated I try to optimize it

domalewski (approved):
yeah looks good 🚀

src/App.tsx
```diff
@@ -15,47 +15,131 @@ export const goodsFromServer = [
...
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, reorderOptions)
+            .map((li) => (
```
pawel-peksa, line 138: I think `li` might not be the best naming here. It does not really say what it is and can be easily confused with html tag element `<li>` 
We have the array of goods here so maybe something like `good` would work?

src/App.tsx
```diff
@@ -15,47 +15,131 @@ export const goodsFromServer = [
...
+              <li data-cy="Good" key={li}>
```
pawel-peksa, line 139: You remembered about the `key` prop, nice! 💪

src/App.tsx
pawel-peksa, line 80: These two functions suggest that maybe we might spilt `reorderOptions` state into two separate things.
At the end, in this two functions we do not change `isReserved` and in the `reverseOrder` we do not change `sortType`

src/App.tsx
```diff
@@ -15,47 +15,131 @@ export const goodsFromServer = [
...
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  // Sort and reverse goods if needed
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
pawel-peksa, line 39: When shipping final version of the code let's remove all `console.log`s unless specified directly in the task description

pawel-peksa (changes requested):
Overall looks nice, good job. 
I've added couple of comments how to improve the code slightly.

PR #1293, opened 2023-09-06
solution

domalewski:
[DEMO LINK](https://domalewski.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,25 +52,25 @@ export function getReorderedGoods(
 }
 
 export const App: React.FC = () => {
-  const [isReversed, setIsReversed] = useState(false);
-  const [sortType, setSortType] = useState(SortType.NONE);
-  const resetButton = sortType !== SortType.NONE || isReversed;
-  const recordedGoods = getReorderedGoods(goodsFromServer, {
-    sortType,
-    isReversed,
+  const [state, setState] = useState({
+    isReversed: false,
+    sortType: SortType.NONE,
   });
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+  };
+
+  const setReverse = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const setReset = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
```
pawel-peksa, line 74: Usually `set` prefix is generally used in state setter. When you define your own function to handle something in response to an event I would propose to use something like:
`handleReverseButton`, `handleReverse`, `handleResetButton`, `handleReset`
It allows to quickly distinguish between function setting the state from `useState` hook and additional custom functions.

src/App.tsx
```diff
@@ -15,48 +16,113 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          onClick={() => handleSort(SortType.ALPHABET)}
```
pawel-peksa, line 81: Nitpick: The logic here is so straight forward that you could even use `setSortType` here:
```suggestion
          onClick={() => setSortType(SortType.ALPHABET)}
```

pawel-peksa (changes requested):
Overall the task done correctly, only couple of minor comments.
Good job 💪

choeqq (approved):

PR #1298, opened 2023-09-06
react_list-of-goods-hooks

annak1104:
https://annak1104.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [
+    sortedGoods,
+    setSortedGoods,
+  ] = useState(goodsFromServer);
```
roman-mirzoian, line 22:
```suggestion
  const [sortedGoods, setSortedGoods] = useState(goodsFromServer);
```

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+  const [isVisibleReser, setIsVisibleReser] = useState(false);
```
roman-mirzoian, line 23: I don't quite understand what you meant by that name
```suggestion
  const [isResetButtonActive, setIsResetButtonActive] = useState(false);
```

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+
+  enum SortType {
+    Alphabet = 'alphabet',
+    Length = 'length',
+    Reverse = 'reverse',
+  }
```
roman-mirzoian, line 29: Things that do not directly affect the component should be taken outside the definition of the component, because they will be created anew each time after each rerender.

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+
+  function handleSort(filter: string) {
```
roman-mirzoian, line 31: You are not filtering, but sorting the list, why is the argument called a `filter`?

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+    let sorted: string[] = [];
```
roman-mirzoian, line 32: Sorted what?

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+      sorted = [...sortedGoods].sort();
+    }
+
+    if (filter === SortType.Length) {
+      sorted = [...sortedGoods].sort((a, b) => a.length - b.length);
+    }
+
+    if (filter === SortType.Reverse) {
+      sorted = [...sortedGoods].reverse();
+    }
```
roman-mirzoian, line 44: You make a copy of the array in three cases, so it is worth doing it once when creating the array at the beginning of this function.

src/App.tsx
```diff
@@ -16,47 +16,93 @@ export const goodsFromServer = [
...
+
+    setSortedGoods(sorted);
+    setIsVisibleReser(true);
+  }
```
roman-mirzoian, line 48: This function should be moved outside the component definition.

src/App.tsx
roman-mirzoian, line 31: Handle sort of what?

roman-mirzoian (changes requested):

BudnikOleksii (changes requested):
The review didn't trigger the tool and didn't change the status in the table

src/App.tsx
Moroz-Dmytro: Not fixed

Moroz-Dmytro (changes requested):
Please, fix the tests, they have failed

src/App.tsx
annak1104: can you help and explain how to do it? because I dont understand. Thanks

Moroz-Dmytro: Sorry, looks, like it was fixed, but better to use `switch/case` here

Moroz-Dmytro (changes requested):
I would approve it, but the tests failed. Fix tests and I'll approve it

PR #1312, opened 2023-09-08
solution

MaciejKujawa:
[DEMO LINK](https://MaciejKujawa.github.io/react_list-of-goods-hooks/)
with problem with deploy


choeqq (approved):
Code looks good, please make sure to deploy your task solution correctly @MaciejKujawa 👀 

We're using comment only when there is some uncommon behaviour inside your code, other than that you don't really have to use them ✅

PR #1327, opened 2023-09-12
add task solution

andrew17752985:
https://andrew17752985.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@
 import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 enum SortType {
   Alphabetically = 'alphabetically',
@@ -71,7 +72,9 @@ export const App: React.FC = () => {
       <div className="buttons">
         <button
           type="button"
-          className={`button is-info ${sortType === SortType.Alphabetically ? '' : 'is-light'}`}
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.Alphabetically,
+          })}
           onClick={() => handleSort(SortType.Alphabetically)}
         >
           Sort alphabetically
@@ -79,7 +82,9 @@ export const App: React.FC = () => {
 
         <button
           type="button"
-          className={`button is-success ${sortType === SortType.Length ? '' : 'is-light'}`}
+          className={classNames('button is-success', {
+            'is-light': sortType !== SortType.Length,
+          })}
           onClick={() => handleSort(SortType.Length)}
         >
           Sort by length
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +21,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+// eslint-disable-next-line max-len
+const stableSort = (array: string[], compareFunction: (a: string, b: string) => number) => {
+  return array
+    .map((item, index) => ({ item, index }))
+    .sort((a, b) => compareFunction(a.item, b.item) || a.index - b.index)
+    .map(({ item }) => item);
+};
```
plodnikate, line 30: Your logic is good, but it can be simplified. Write one universal simple function that will return sorted data, based on sortType and isReversed. You need to use it to have sortedGoods in all places.

src/App.tsx
```diff
@@ -15,48 +21,112 @@ export const goodsFromServer = [
...
+
+  const handleSort = (type: SortType) => {
+    let sortedGoods = [...goods];
+
+    if (type === SortType.Alphabetically) {
+      sortedGoods = stableSort([...goods], (a, b) => a.localeCompare(b));
+    } else if (type === SortType.Length) {
+      sortedGoods = stableSort(
+        [...goods],
+        (a, b) => a.length - b.length || goods.indexOf(a) - goods.indexOf(b),
```
plodnikate, line 47: It's enough to do sort like this `copyGoods.sort((a, b) => a.length - b.length);`.

plodnikate (changes requested):
Good job! But simplified it as much as possible!

src/App.tsx
```diff
@@ -16,47 +23,93 @@ export const goodsFromServer = [
...
+
+  const isResetVisible = sortType !== SortType.Default || !ascending;
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sortType === SortType.Alphabetically ? '' : 'is-light'}`}
```
etojeDenys, line 74: let's use classNames for such cases

etojeDenys (changes requested):
image

image

reverse button should also have is-light class

src/App.tsx
```diff
@@ -16,47 +24,99 @@ export const goodsFromServer = [
...
+  const [ascending, setAscending] = useState<boolean>(true);
+
+  const sortGoods = (items: string[], type: SortType, isReversed: boolean): string[] => {
+    const sortedItems = [...items];
+
+    if (type === SortType.Alphabetically) {
+      sortedItems.sort((a, b) => b.localeCompare(a));
+    } else if (type === SortType.Length) {
+      sortedItems.sort((a, b) => a.length - b.length);
+    }
```
sTorba24, line 38: use switch case instead of it

src/App.tsx
```diff
@@ -16,47 +24,99 @@ export const goodsFromServer = [
...
         <button
           type="button"
-          className="button is-warning is-light"
+          className={classNames('button is-warning', {
+            'is-light': ascending,
+          })}
+          onClick={handleReverse}
         >
-          Reverse
+          Реверс
```
sTorba24, line 100: why do you need this translation ?)

sTorba24 (approved):

PR #1329, opened 2023-09-16
add task solution

Anmonnn:
[DEMO LINK](https://Anmonnn.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+  const array = [...goods];
+
+  array.sort((value1, value2) => {
+    if (howSort === SortType.NAME) {
+      return value1.localeCompare(value2);
+    }
+
+    if (howSort === SortType.LENGTH) {
+      return value1.length - value2.length;
+    }
```
Anastasiia-Svintsova, line 37: You can use the switch/case instead. Let's imagine that there will be more than 2 options

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
 export const App: React.FC = () => {
+  const [sort, setSort] = useState('');
+  const [isReversed, setIsReversed] = useState(false);
+  const visibleGoods = sortArray(
+    goodsFromServer,
+    sort as SortType,
+    isReversed,
+  );
+
+  function resetAll() {
```
Anastasiia-Svintsova, line 58: It's better to use arrow functions for better code readability

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    array.reverse();
+  }
+
+  return array;
+}
+
 export const App: React.FC = () => {
+  const [sort, setSort] = useState('');
```
Anastasiia-Svintsova, line 50: You can add an empty string to the sort type as a default type. For example
```suggestion
  const [sort, setSort] = useState(SortType.NONE);
```

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(false);
+  const visibleGoods = sortArray(
+    goodsFromServer,
+    sort as SortType,
```
Anastasiia-Svintsova, line 54: Then you'll be able to simplify it
```suggestion
    sort,
```

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+          ? (
+            <button
+              type="button"
+              className={`button is-danger ${sort !== '' || isReversed ? null : 'is-light'}`}
+              onClick={resetAll}
+            >
+              Reset
+            </button>
+          )
+          : null}
```
Anastasiia-Svintsova, line 101: Seems that the ternary is redundant here. Use `&&` instead

src/App.tsx
```diff
@@ -15,48 +15,97 @@ export const goodsFromServer = [
...
+    setIsReversed(false);
+  }
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sort === SortType.NAME ? null : 'is-light'}`}
```
Anastasiia-Svintsova, line 69: You can use a classnames library as well

Anastasiia-Svintsova (changes requested):
Great start! Check my comments, please

src/App.tsx
volodymyr-soltys97: Still not fixed

src/App.tsx
```diff
@@ -15,48 +15,96 @@ export const goodsFromServer = [
...
+              className={`button is-danger ${sort !== SortType.NONE || isReversed ? null : 'is-light'}`}
+              onClick={() => {
+                setSort(SortType.NONE);
+                setIsReversed(false);
+              }}
+            >
+              Reset
+            </button>
+          )
+          : null}
```
volodymyr-soltys97, line 100:
```suggestion
          )}
```

src/App.tsx
```diff
@@ -15,48 +15,96 @@ export const goodsFromServer = [
...
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {sort !== SortType.NONE || isReversed
+          ? (
```
volodymyr-soltys97, line 88:
```suggestion
          && (
```

src/App.tsx
```diff
@@ -15,48 +15,96 @@ export const goodsFromServer = [
...
+          className={`button is-success ${sort === SortType.LENGTH ? null : 'is-light'}`}
+          onClick={() => setSort(SortType.LENGTH)}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`button is-warning ${isReversed === true ? null : 'is-light'}`}
```
volodymyr-soltys97, line 81: Consider will be better if you use classnames library
```suggestion
          className={`button is-warning ${isReversed ? null : 'is-light'}`}
```

volodymyr-soltys97 (changes requested):
Let's improve your code

src/App.tsx
```diff
@@ -15,48 +16,111 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  NAME = 'name',
+  LENGTH = 'length',
+  NONE = 'none',
+}
+
+function sortArray(
+  goods: string[],
+  howSort: SortType,
```
VitaliyBodnar, line 27: `howSort` it's not a correct name, because it's sounds as question with some actions
it's better to rename it to `type` or `sortType`

src/App.tsx
```diff
@@ -15,48 +16,111 @@ export const goodsFromServer = [
...
+            <button
+              type="button"
+              className={classNames(
+                'button',
+                'is-danger',
+                { 'is-light': !(sort !== SortType.NONE || isReversed) },
+              )}
+              onClick={() => {
+                setSort(SortType.NONE);
+                setIsReversed(false);
```
VitaliyBodnar, line 111: it is bad practise to make anonymous arrow functions in jsx.

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1330, opened 2023-09-16
add task solution  💙💛

denys2:
[LINK](https://denys2.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +15,122 @@ export const goodsFromServer = [
...
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
```
VitaliyBodnar, line 41: it's better to use switch case statement here

src/App.tsx
```diff
@@ -15,47 +15,122 @@ export const goodsFromServer = [
...
         >
           Sort by length
-        </button>
 
+        </button>
         <button
           type="button"
-          className="button is-warning is-light"
+          className={
+            isReversed === true
```
VitaliyBodnar, line 101: it's better to use classname library for making dynamic classNames 
```suggestion
            isReversed
```

src/App.tsx
```diff
@@ -15,47 +15,122 @@ export const goodsFromServer = [
...
-          Reset
         </button>
+        {(sortType || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={resetGoods}
+          >
+            Reset
+
```
VitaliyBodnar, line 117:
```suggestion
```

src/App.tsx
```diff
@@ -15,47 +15,122 @@ export const goodsFromServer = [
...
 
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, { sortType, isReversed }).map(
```
VitaliyBodnar, line 124: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):

src/App.tsx
lerastarynets: not fixed

src/App.tsx
```diff
@@ -15,48 +15,121 @@ export const goodsFromServer = [
...
       <ul>
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getReorderedGoods(goodsFromServer, { sortType, isReversed }).map(
```
lerastarynets, line 126: i think you were previously asked to move `getReorderedGoods` to separate const/var

lerastarynets (changes requested):
Pls fix prev comments

denys2:
please update the comments, because I can't think, since the order of the code was changed in 
solution 2

src/App.tsx
```diff
@@ -15,48 +15,121 @@ export const goodsFromServer = [
...
         >
           Sort by length
-        </button>
 
+        </button>
         <button
           type="button"
-          className="button is-warning is-light"
+          className={
+            isReversed === true
```
sTorba24, line 105:
```suggestion
            isReversed
```

src/App.tsx
sTorba24: image
 
 
 move it to variable

sTorba24 (changes requested):
Well done! Some fixes needs to be done

also add styles to buttons(as it is on working version)
image

it should be different
image

sTorba24 (approved):
Well done!

PR #1348, opened 2023-10-02
add task solution

kostiantin-hudzenko:
[DEMO LINK](https://kostiantin-hudzenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,101 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField !== '' || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setIsReversed(false);
+              setSortField('');
+            }}
```
VitaliyBondarenko1982, line 103: make handler in component and pass reference here
```
const handleReset = () => {
  setIsReversed(false);
  setSortField('');
}
```
```suggestion
            onClick={handleReset}
```

src/App.tsx
```diff
@@ -15,48 +16,101 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField !== '' || isReversed) && (
```
VitaliyBondarenko1982, line 96: better to make separate enum field for default sort.

```
enum SortType {
  none = 'none',
  alphabet = 'alphabet',
  length = 'length',
}
```

src/App.tsx
```diff
@@ -15,48 +16,101 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn('button is-info', {
+            'is-light': sortField !== SortType.alphabet,
+          })}
+          onClick={() => setSortField(SortType.alphabet)}
```
VitaliyBondarenko1982, line 71: for both sort buttons make one handler factory which will accept SortType and return arrow function
```
const hanldeSort = (sort: SortType) => () => setSortField(sort);
```
```suggestion
          onClick={handleSort(SortType.alphabet)}
```

src/App.tsx
```diff
@@ -15,48 +16,101 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={cn('button is-success', {
+            'is-light': sortField !== SortType.length,
+          })}
+          onClick={() => setSortField(SortType.length)}
```
VitaliyBondarenko1982, line 81: same as for reset make handler and pass reference here.
it is bad practise to make anonymous arrow functions in jsx.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check a few minor comments.

IvanFesenko (changes requested):
You should check demo limk before request review
2023-10-02_15h32_51

PR #1370, opened 2023-10-16
add task solution

MishaLehotskyi:
[DEMO LINK](https://MishaLehotskyi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,8 +3,14 @@ import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+enum SortType {
+  name = 'name',
+  length = 'length',
+  none = '',
+}
+
 type SortParams = {
-  property: string,
+  property: SortType,
   reverse: boolean,
 };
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+  reverse: boolean,
+};
+
+function getGoods(goods: string[], sortParams: SortParams) {
+  const goodsCopy = [...goods];
+
+  if (sortParams.property) {
+    goodsCopy.sort((good1, good2) => {
+      switch (sortParams.property) {
+        case 'name':
```
VitaliyBodnar, line 17: please move 'name' and 'length' to the `SortType` enum as it will make a code cleaner

src/App.tsx
```diff
@@ -16,46 +45,76 @@ export const goodsFromServer = [
...
           type="button"
-          className="button is-info is-light"
+          className={cn(
+            'button is-info',
+            { 'is-light': sortParams.property !== 'name' },
+          )}
+          onClick={() => setSortParams({
+            ...sortParams,
+            property: 'name',
+          })}
```
VitaliyBodnar, line 66: it is bad practise to make anonymous arrow functions in jsx.

VitaliyBodnar (changes requested):

src/App.tsx
anastasiiavorobiova: Not fixed from the previous review

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #1376, opened 2023-10-18
solution

Masha-Cactus:


src/App.tsx
```diff
@@ -15,48 +16,100 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
+  byAlphabet = 'Sort alphabetically',
+  byLength = 'Sort by length',
```
VitaliyBodnar, line 21: enum's elements should be in CamelCase or UPPER_CASE

src/App.tsx
```diff
@@ -15,48 +16,100 @@ export const goodsFromServer = [
...
+  setSortField: React.Dispatch<React.SetStateAction<string>>,
+  setReverse: React.Dispatch<React.SetStateAction<boolean>>,
+) {
+  setSortField('');
+  setReverse(false);
+}
+
+export const App : React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reverse, setReverse] = useState(false);
```
VitaliyBodnar, line 61: reverse is sounds as a functionName but it is a boolean property, so I suggest you to rename it to the` [isReversed, setIsReversed]`

src/App.tsx
```diff
@@ -15,48 +16,100 @@ export const goodsFromServer = [
...
+    : preparedGoods;
+}
+
+function reset(
+  setSortField: React.Dispatch<React.SetStateAction<string>>,
+  setReverse: React.Dispatch<React.SetStateAction<boolean>>,
+) {
+  setSortField('');
+  setReverse(false);
+}
```
VitaliyBodnar, line 57: don't see any sense to use these callbacks as a parameters of function, because you can just execute these methods from this function 
```suggestion
function reset() {
  setSortField('');
  setReverse(false);
}
```

VitaliyBodnar (changes requested):

src/App.tsx
Masha-Cactus: okey, but now elements are in CamelCase)

src/App.tsx
Masha-Cactus: no, in this case they are invisible)

nazarmatsevych (approved):
Well done!

PR #1383, opened 2023-10-18
Solution

adioskv:
[**DEMO LINK**](https://adioskv.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import React, { useState } from 'react';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +16,110 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType>(SortType.Null);
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+
+  function reset(): void {
+    setSortType(SortType.Null);
+    setIsReversed(false);
+  }
+
```
OlhaKoziuk, line 69: I'm not sure? but as i remember, we can not write type void if the function return nothing

src/App.tsx
```diff
@@ -15,48 +16,110 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {sortedGoods.map((elem: string) => (
+          <li data-cy="Good" key={elem}>
+            {elem}
+          </li>
```
OlhaKoziuk, line 121: maybe you can use more understandable names, like good or product

OlhaKoziuk (approved):
Great job! Keep going!

src/App.tsx
adioskv: we can, but i want to

src/App.tsx
adioskv: okay, thank you !

src/App.tsx
```diff
@@ -15,48 +16,110 @@ export const goodsFromServer = [
...
+];
+
+const sortByParams = (
+  toSortArray: string[],
+  sortType: SortType,
+  isReversed: boolean,
+): string[] => {
+  let array: string[] = [...toSortArray];
+
+  array = !SortType.Null && array.sort((a: string, b: string) => {
```
vasylmiskiv, line 44:
```suggestion
if (sortType) {
  array.sort((product: string, product: string) => {
```

src/App.tsx
```diff
@@ -15,48 +16,110 @@ export const goodsFromServer = [
...
+    switch (sortType) {
+      case SortType.Name:
+        return a.localeCompare(b);
+
+      case SortType.Length:
+        return a.length - b.length;
+
+      case SortType.Null:
+      default:
+        return 0;
```
vasylmiskiv, line 54:
```suggestion
      default:
        return 0;
```
The default case does not require specific orders

vasylmiskiv (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,110 @@ export const goodsFromServer = [
...
+}
+
+const buttons: Buttons[] = [
+  { sortType: SortType.Name, buttonStyle: 'is-info' },
+  { sortType: SortType.Length, buttonStyle: 'is-success' },
+  { sortType: SortType.Reverse, buttonStyle: 'is-warning' },
+];
+
+const sortByParams = (
+  toSortArray: string[],
```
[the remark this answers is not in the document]
adioskv: `toSortArray` means array, that we need to sort, so we can reuse our function without naming "goods" or "products", we can put anything except "goods", so that naming doesn't fit correctly on my opinion

PR #1388, opened 2023-10-18
Solution

YaroslavYarynych:
[DEMO LINK](https://YaroslavYarynych.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,12 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+enum Buttons {
+  SORT__ALPHBETICALLY = 'sort__alphabetically',
+  SORT__BY__LENGTH = 'sort__by__length',
+}
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,13 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+enum Buttons {
```
o-drozzdyk, line 6: You should also add default value for sorting here and then use it in useState instead of null

src/App.tsx
```diff
@@ -15,48 +21,106 @@ export const goodsFromServer = [
...
+  isReversed: boolean,
+) {
+  const copyGoods = [...goods];
+
+  if (sortType) {
+    copyGoods.sort((good1, good2) => {
+      switch (sortType) {
+        case 'sort__alphabetically':
+          return good1.localeCompare(good2);
+        case 'sort__by__length':
```
o-drozzdyk, line 36: And here Buttons.SORT__BY__LENGTH

src/App.tsx
```diff
@@ -15,48 +21,106 @@ export const goodsFromServer = [
...
+  );
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn(
+            'button',
```
o-drozzdyk, line 69: You can write constant classes in 1 line like 'button is-info'

o-drozzdyk (changes requested):
Good job! but fix some moments please)

src/App.tsx
```diff
@@ -15,48 +22,104 @@ export const goodsFromServer = [
...
+
+  if (sortType) {
+    copyGoods.sort((good1, good2) => {
+      switch (sortType) {
+        case Buttons.SORT__ALPHBETICALLY:
+          return good1.localeCompare(good2);
+        case Buttons.SORT__BY__LENGTH:
+          return good1.length - good2.length;
+
+        default: return 0;
```
polosanya, line 40: Would be great to move default case out of the sort callback to avoid extra sort calls

src/App.tsx
```diff
@@ -15,48 +22,104 @@ export const goodsFromServer = [
...
+
+  return isReversed ? copyGoods.reverse() : copyGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setsortType] = useState<Buttons>(Buttons.SORT__BY__DEFAULT);
+  const [isReversed, setIsReversed] = useState(false);
+  const preparedGoods = getPreparedGoods(
+    goodsFromServer, sortType, isReversed,
+  );
```
polosanya, line 53:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
  
  const preparedGoods = getPreparedGoods(
    goodsFromServer, sortType, isReversed,
  );
```

src/App.tsx
```diff
@@ -15,48 +22,104 @@ export const goodsFromServer = [
...
+
+  const handleReset = () => {
+    setsortType(Buttons.SORT__BY__DEFAULT);
+    setIsReversed(false);
+  };
+
+  const resetCondition = (
```
polosanya, line 60:
```suggestion
  const isResetVisible = (
```
image

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #1397, opened 2023-10-18
Develop

BetterSol:
https://BetterSol.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,11 +20,12 @@ export const goodsFromServer = [
 export enum SortType {
   Alphabet = 'abc',
   Length = 'length',
+  Empty = '',
 }
 
 function getSortedGoods(
   goods: string[],
-  sortField: SortType | '',
+  sortField: SortType,
   isReversed: boolean,
 ): string[] {
   const sortedGoodsCopy = [...goods];
@@ -49,7 +50,7 @@ function getSortedGoods(
 }
 
 export const App: React.FC = () => {
-  const [sortField, setSortField] = useState<SortType | ''>('');
+  const [sortField, setSortField] = useState<SortType>(SortType.Empty);
   const [isReversed, setIsReversed] = useState(false);
   const sortedGoods = getSortedGoods(goodsFromServer, sortField, isReversed);
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
...
+    sortedGoodsCopy.sort((good1, good2) => {
+      return good1.localeCompare(good2);
+    });
+  }
+
+  if (sortField === SortType.Length) {
+    sortedGoodsCopy.sort((good1, good2) => {
+      return good1.length - good2.length;
+    });
+  }
```
koros-rk, line 43: Its better to use swich-case

src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+export enum SortType {
+  Alphabet = 'abc',
+  Length = 'length',
+  Default = '',
+}
```
koros-rk, line 24: You can remove enum values as you don't use them into anywhere else

```suggestion
export enum SortType {
  Alphabet,
  Length,
  Default,
}
```

koros-rk (changes requested):
Good job )

src/App.tsx
BetterSol: agree)

src/App.tsx
BetterSol: I tried to do it with `switch-case`, but I prefer to write it with `if`, because in this case we have only one return of the default meaning instead of two (the default after switch-case and the other one after reversed condition). That is why I decided to stay with if-statement.

src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn(
+            'button',
+            'is-info',
```
StasSokolov, line 64:
```suggestion
            'button is-info',
```

can be simplify, check all places

src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
...
-        >
-          Reset
-        </button>
+        {(sortField !== SortType.Default || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortType.Default);
+              setIsReversed(false);
```
StasSokolov, line 108: will be better create separated function for this logic like `reset`

StasSokolov (changes requested):
Almost done, only few minor comments from me!

src/App.tsx
BetterSol: done)

src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
...
+): string[] {
+  const sortedGoodsCopy = [...goods];
+
+  if (sortField === SortType.Alphabet) {
+    sortedGoodsCopy.sort((good1, good2) => {
+      return good1.localeCompare(good2);
+    });
+  }
+
+  if (sortField === SortType.Length) {
```
polosanya, line 39: Would be great to use switch...case here as you have a limited amount of sortFields

src/App.tsx
```diff
@@ -15,48 +17,108 @@ export const goodsFromServer = [
...
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn(
+            'button', 'is-warning',
+            { 'is-light': !isReversed },
+          )}
+          onClick={() => {
+            setIsReversed(!isReversed);
```
polosanya, line 99: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

polosanya (approved):
👍

PR #1398, opened 2023-10-18
add task solution

mpohorenyi:
[DEMO LINK](https://mpohorenyi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { v4 as uuidv4 } from 'uuid';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +17,129 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map((good) => (
+          <li data-cy="Good" key={good.id}>
+            {good.name}
+          </li>
+        ))}
```
koros-rk, line 142: its better to use object destructuring

```suggestion
        {visibleGoods.map(({ id, name }) => (
          <li data-cy="Good" key={id}>
            {name}
          </li>
        ))}
```

src/App.tsx
```diff
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { v4 as uuidv4 } from 'uuid';
```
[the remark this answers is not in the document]
mpohorenyi: Not working

src/App.tsx
mpohorenyi: I don't think it's necessary

src/App.tsx
```diff
@@ -15,48 +17,129 @@ export const goodsFromServer = [
...
+  if (reverse) {
+    preparedGoods = preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>({
+    sortType: null,
```
vasylmiskiv, line 70: Use an empty string for example and add it to your `enum` and separate reverse to another state with boolean value

vasylmiskiv (approved):

src/App.tsx
vasylmiskiv: Also don't forget about Keep it simple principle and try to simplify the code as much as possible

PR #1401, opened 2023-10-18
task solution

dolphinnq:
[DEMO LINK](https://dolphinnq.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || isReversed) && (
+          <button
```
dsfreedom, line 104: I really don`t get why, but your reset button goes fully red if you click only on reverse and it does not fully match the working version. Also it would be cool to create seperate function for reset button as we could have more things to reset.
image

src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
```
dsfreedom, line 19: You could've add the initial state as None  = '' here and use it later as default for sortField.

dsfreedom (approved):

src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className={cn(
+              'button',
+              'is-danger',
+              { 'is-light': sortField !== '' },
+            )}
+            onClick={() => {
+              setSortField('');
+              setIsReversed(false);
```
vasylmiskiv, line 113: Would be nice to create a separate handler function for reset as @dsfreedom mentioned above

vasylmiskiv (approved):
Well done 🚀

PR #1402, opened 2023-10-18
add task solution

dsfreedom:
[DEMO LINK](https://dsfreedom.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+
+import { SortType, SortingParams } from './Types/SortingParams';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


tetyana-pol (approved):

src/App.tsx
```diff
@@ -15,48 +18,112 @@ export const goodsFromServer = [
...
+        {isResetVisible
+        && (
+          <button
+            type="button"
+            onClick={resetSorting}
+            className="button is-danger is-light"
+          >
+            Reset
+          </button>
+        )}
```
StasSokolov, line 118:
```suggestion
        {isResetVisible && (
          <button
            type="button"
            onClick={resetSorting}
            className="button is-danger is-light"
          >
            Reset
          </button>
        )}
```

check formatting please

StasSokolov (changes requested):
Almost done, let's improve your code a bit more!

src/App.tsx
```diff
@@ -15,48 +18,118 @@ export const goodsFromServer = [
...
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.None);
+  const [isReversed, setReversed] = useState<boolean>(false);
+  const [isResetVisible, setIsResetVisible] = useState<boolean>(false);
```
superpooperxxx, line 57: This is an overkill, this value should be calculated from `sortField` and `isReversed`, but should not be saved to state, because this will trigger extra renders

src/App.tsx
```diff
@@ -15,48 +18,118 @@ export const goodsFromServer = [
...
+      setReversed(false);
+    } else if (reverse && sortField !== SortType.None) {
+      setReversed(false);
+    } else if (!reverse && sortField !== SortType.None) {
+      setReversed(true);
+    } else if (!reverse && sortField === SortType.None) {
+      setIsResetVisible(true);
+      setReversed(true);
+    }
+  }
```
superpooperxxx, line 76: After you get rid of `isResetVisible` state, you can get rid of this

superpooperxxx (changes requested):
I like your solution, but we can improve it

PR #1404, opened 2023-10-19
solution

MVoutch:
[DEMO LINK](https://MVoutch.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { SortBy } from './types/SortBy';
+import { Buttons } from './components/Buttons';
+import { Products } from './components/Products/Products';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/components/Products/Products.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+interface Props {
+  goods: string[]
+}
+export const Products: React.FC<Props> = ({ goods }) => (
+  <ul>
+    <ul>
+      {goods.map((x: string) => (
+        <li data-cy="Good">{x}</li>
+      ))}
```
vasylmiskiv, line 11:
```suggestion
      {goods.map((good: string) => (
        <li data-cy="Good" key={good}>{good}</li>
      ))}
```

src/App.tsx
```diff
@@ -16,48 +19,47 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [isReversed, setIsReversed] = useState(false);
+  const goods: string[] = [...goodsFromServer];
```
vasylmiskiv, line 24: copy of the array of goods should be inside the separate function and return the sorted goods instead of writing like so

vasylmiskiv (approved):
Everything looks almost fine but look at the comments

PR #1408, opened 2023-10-19
add task solution

markiuan:
[DEMO LINK](https://markiuan.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable max-len */
-/* eslint-disable @typescript-eslint/quotes */
 import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
@@ -30,27 +28,25 @@ const sortGoods = (goodsArray: string[], sortType: SortType,
   switch (sortType) {
     case SortType.Alphabetical:
       sortedGoods = sortedGoods.sort();
-      if (isReverse) {
-        sortedGoods.reverse();
-      }
-
-      return sortedGoods;
-    default:
+      break;
+    case SortType.Length:
       sortedGoods = [...goodsFromServer].sort((a, b) => {
-        const lengthComparison = a.length - b.length;
-
-        if (lengthComparison === 0) {
+        if (a.length - b.length === 0) {
           return 1;
         }
 
-        return lengthComparison;
+        return a.length - b.length;
       });
-      if (isReverse) {
-        sortedGoods.reverse();
-      }
-
-      return sortedGoods;
+      break;
+    default:
+      return [...goodsFromServer];
   }
+
+  if (isReverse) {
+    sortedGoods.reverse();
+  }
+
+  return sortedGoods;
 };
 
 export const App: React.FC = () => {
@@ -83,7 +79,6 @@ export const App: React.FC = () => {
       <div className="buttons">
         <button
           type="button"
-          // eslint-disable-next-line no-sequences, @typescript-eslint/object-curly-spacing
           className={`button is-info ${(sortType === SortType.Alphabetical) ? '' : 'is-light'}`}
           onClick={() => handleSort(SortType.Alphabetical)}
         >


src/App.tsx
```diff
@@ -15,48 +15,107 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {goods.join(',') !== goodsFromServer.join(',') && (
```
etojeDenys, line 102: you can just check if `sortType` or `isReverse` is set

src/App.tsx
```diff
@@ -15,48 +15,107 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goods.map(good => (
+          <li data-cy="Good">
+            {good}
+          </li>
```
etojeDenys, line 117: do not forget to add key prop

src/App.tsx
```diff
@@ -15,48 +15,107 @@ export const goodsFromServer = [
...
+      if (isReverse) {
+        setGoods([...sortedGoods].reverse());
+      } else {
+        setGoods(sortedGoods);
+      }
+
+      setSortType(SortType.Length);
+    }
+  };
+
```
etojeDenys, line 61: try to combine these functions in one separate function that will return the sorted array of todos

it will improve at least the readability

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,107 @@ export const goodsFromServer = [
...
+  Length = 'Length',
+}
+
+const sortGoods = (goodsArray: string[], sortType: SortType,
+  isReverse: boolean): string[] => {
+  let sortedGoods = [...goodsArray];
+
+  if (sortType === SortType.Alphabetical) {
+    sortedGoods = sortedGoods.sort();
+  } else if (sortType === SortType.Length) {
```
OleziO, line 30: Use switch case constructor to optimize code

OleziO (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,112 @@ export const goodsFromServer = [
...
+    setIsReverse(false);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${(sortType === SortType.Alphabetical) ? '' : 'is-light'}`}
```
etojeDenys, line 84:
```suggestion
          className={'button is-info', {"is-light": sortType === SortType.Alphabetical}}
```

etojeDenys (changes requested):

lerastarynets (changes requested):
.

src/App.tsx
```diff
@@ -15,48 +18,113 @@ export const goodsFromServer = [
...
+const sortGoods = (goodsArray: string[], sortType: SortType,
+  isReverse: boolean): string[] => {
+  let sortedGoods = [...goodsArray];
+
+  switch (sortType) {
+    case SortType.Alphabetical:
+      sortedGoods = sortedGoods.sort();
+      if (isReverse) {
+        sortedGoods.reverse();
+      }
```
VitaliyBodnar, line 35: this code should not be duplicated in every case, is should be executed only once on after the switch block 
```
if (isReverse) {
     sortedGoods.reverse();
 }
```

src/App.tsx
```diff
@@ -15,48 +18,113 @@ export const goodsFromServer = [
...
+
+      return sortedGoods;
+    default:
+      sortedGoods = [...goodsFromServer].sort((a, b) => {
```
VitaliyBodnar, line 39: - by default it should return just [...goodsFromServer], because it's a default case 
- this sorting logic move to the separate case (Length)

src/App.tsx
VitaliyBodnar: it's better to use classname library for making a dynamic classnames

src/App.tsx
```diff
@@ -1,6 +1,9 @@
-import React from 'react';
+/* eslint-disable max-len */
+/* eslint-disable @typescript-eslint/quotes */
```
VitaliyBodnar, line 2: delete these eslint-disables and fix the linter issues

src/App.tsx
```diff
@@ -15,48 +18,113 @@ export const goodsFromServer = [
...
+      return sortedGoods;
+    default:
+      sortedGoods = [...goodsFromServer].sort((a, b) => {
+        const lengthComparison = a.length - b.length;
+
+        if (lengthComparison === 0) {
+          return 1;
+        }
+
+        return lengthComparison;
```
VitaliyBodnar, line 46: can be less
```suggestion
        return a.length - b.length;
```

VitaliyBodnar (changes requested):

nazarmatsevych (approved):
👍

PR #1410, opened 2023-10-19
tryTS

Volodymyr-Kustikov:


Andriy-Shymkiv (changes requested):
your tests is falling, please fix them and request a review again

src/App.tsx
```diff
@@ -1,6 +1,15 @@
...
 import './App.scss';
+import cn from 'classnames';
+
+interface Methods {
+  sortField: string;
+  isReversed: boolean;
+}
+
+const alphabeticallySort = 'alphabet';
+const lengthSort = 'length';
```
VitaliyBodnar, line 12: it's better to use UPPER_CASE for static constants

src/App.tsx
```diff
@@ -15,48 +24,96 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField || isReversed) && (
+
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField('');
+              setIsReversed(false);
```
VitaliyBodnar, line 103: it is bad practise to make anonymous arrow functions in jsx.

VitaliyBodnar (changes requested):

PR #1413, opened 2023-10-24
react list of goods hooks

whiovt:


anastasiia-tilikina (changes requested):
please provide demo link

PR #1434, opened 2023-11-04
Develop

mukutiuk:
https://mukutiuk.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -78,9 +78,9 @@ export const App: React.FC = () => {
         <button
           type="button"
           className={`button is-warning ${cn({
-            'is-light': reverse === false,
+            'is-light': !reverse,
           })}`}
-          onClick={() => setReverse(!reverse)}
+          onClick={() => setReverse((prevReverse) => !prevReverse)}
         >
           Reverse
         </button>


src/App.tsx
```diff
@@ -15,47 +16,95 @@ export const goodsFromServer = [
...
+
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`button is-warning ${cn({
+            'is-light': reverse === false,
```
SanyaBratashchuk, line 81:
```suggestion
            'is-light': !reverse,
```

src/App.tsx
```diff
@@ -15,47 +16,95 @@ export const goodsFromServer = [
...
+          })}`}
+          onClick={() => setReverse(!reverse)}
```
SanyaBratashchuk, line 83:
```suggestion
          onClick={() => setReverse((prevReverse) => !prevReverse)}
```

SanyaBratashchuk: It is safer to pass stateUpdater callback if new state value is based on previous one

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1451, opened 2023-11-20
Develop

ArtemSydun:
[DEMO LINK](https://artemsydun.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,48 +17,107 @@ export const goodsFromServer = [
...
+
+  if (sortField) {
+    preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.ALPHABETIC:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+        default:
+          return 0;
```
polosanya, line 45: It would be great to move switch...case out of the sort callback so there is no need to extra call it on default case

src/App.tsx
```diff
@@ -15,48 +17,107 @@ export const goodsFromServer = [
...
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortType.NONE);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const readyGoods = getPreparedGoods(goodsFromServer,
+    { sortField, isReversed });
+
+  const handleReverse = () => setIsReversed(!isReversed);
```
polosanya, line 64: It is better to toggle the previous value using callback props
https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/App.tsx
```diff
@@ -15,48 +17,107 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortField !== 0) && (
```
polosanya, line 104: Compare the sortField with the Enum value, not just 0

src/App.tsx
```diff
@@ -15,48 +17,107 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE,
+  ALPHABETIC,
+  LENGTH,
```
polosanya, line 23: There is no need to use upper case for Enum values

polosanya (changes requested):
Almost done!

src/App.tsx
```diff
@@ -15,48 +17,109 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(isReversed || sortField !== SortType.None) && (
```
BudnikOleksii, line 106: Create a variable for current condition to avoid combine different logical operators

BudnikOleksii (approved):
Good job 👍

PR #1452, opened 2023-11-21
Develop

vanya-kalyenichenko:
[DEMO LINK](https://vanya-kalyenichenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
+import cn from 'classnames';
 import './App.scss';
+import { useState } from 'react';
+import { ComponentList } from './Components/ComponentList/ComponentList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -15,49 +17,95 @@ export const goodsFromServer = [
...
+    });
+  }
+
+  if (!sortField) {
+    visibleGoods = [...goodsFromServer];
+  }
+
+  if (reverse) {
+    visibleGoods.reverse();
+  }
```
etojeDenys, line 60: let's move this logic to some helper function, and just call it here

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,49 +17,101 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
+  abc = 'abc',
+  length = 'length',
+  reset = '',
```
OleziO, line 23:
```suggestion
  Abc = 'abc',
  Length = 'length',
  Reset = '',
```

OleziO (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1456, opened 2023-11-22
result

errisi:
[DEMO LINK](https://errisi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
+          return good1.localeCompare(good2);
+
+        case SortType.SORT_FILED_LENGTH:
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+  }
```
ArtemSydun, line 44: It would be great to move switch...case out of the sort callback so there is no need to extra call it on default case

src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  SORT_FILED_DEFAULT = '',
+  SORT_FILED_NAME = 'name',
+  SORT_FILED_LENGTH = 'length',
```
ArtemSydun, line 22: There is no need to use upper case for Enum values

ArtemSydun (changes requested):
check the comments

ArtemSydun (approved):
switch case not fixed but let mentor check it

src/App.tsx
```diff
@@ -15,48 +16,119 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || isReversed) && (
```
BudnikOleksii, line 112: Create a variable for `sortField || isReversed` condition to avoid combining different logical operators

BudnikOleksii (approved):
Good work 👍

PR #1466, opened 2023-11-27
add task solution

DamianAdamczewski:
[DEMO LINK](https://DamianAdamczewski.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
+            : good1.length - good2.length;
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (isReverse) {
+    preparedGoods = [...goodsFromServer].reverse();
+  }
```
volodymyr-soltys97, line 54:
```suggestion
  if (isReverse) {
    preparedGoods.reverse();
  }
```

src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={cn({
+            button: true,
+            'is-warning': true,
+            'is-light': !isReverse,
+          })}
+          onClick={() => setIsReverse(!isReverse)}
```
volodymyr-soltys97, line 100: Move this logic from jsx to the helper function and use it here

src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(isReverse || !!sortBy) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortBy('');
+              setIsReverse(false);
+            }}
```
volodymyr-soltys97, line 111: The same as above

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
```diff
@@ -15,48 +16,114 @@ export const goodsFromServer = [
...
+}
+
 export const App: React.FC = () => {
+  const [sortBy, setSortBy] = useState('');
+  const [isReverse, setIsReverse] = useState(false);
+  const goods = [...goodsFromServer];
+
+  const preparedGoods = getPreparedGoods(goods, sortBy, isReverse);
+
+  const handleIsReverse = () => setIsReverse(!isReverse);
```
yevhenii-pyl, line 66: When you plan to use your previous state to change next state - it's always preferable to write it that way: 

`const handleIsReverse = () => setIsReverse(prevReversed => !prevReversed);`

This may save you from some bugs due to useState hook nature.

yevhenii-pyl (changes requested):
Just one little fix and we are good to go!
It's really important for you to remember that pattern with useState :)

yevhenii-pyl (approved):
Now we are talking 👍

PR #1473, opened 2023-11-30
solution

emilionila:
[DEMO LINK](https://emilionila.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


BoblykNazar (approved):
Good job!

src/App.tsx
```diff
@@ -15,48 +17,99 @@ export const goodsFromServer = [
...
+
+  if (reversed) {
+    sortedProducts.reverse();
+  }
+
+  return sortedProducts;
+};
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<string>('');
```
BudnikOleksii, line 57: `sortField` should have a type of `SortFieldType`

src/App.tsx
```diff
@@ -15,48 +17,99 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+
+  const visibleProducts
+    = sortProducts(goodsFromServer, sortField as SortFieldType, isReversed);
```
BudnikOleksii, line 61: And in this case you don't need to cheat with `as`

BudnikOleksii (changes requested):
Good work 👍 
Just one moment that `sortField` type should be fixed

src/App.tsx
```diff
@@ -15,48 +17,99 @@ export const goodsFromServer = [
...
+        {sortField || isReversed ? (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={resetSorting}
+          >
+            Reset
+          </button>
+        )
+          : null}
```
BudnikOleksii, line 106:
```suggestion
        {sortField || isReversed && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={resetSorting}
          >
            Reset
          </button>
        )}
```
You don't ternary operator in this case, also, it's better to create separate variable for `sortField || isReversed` condition to avoid mixing different logical operators

BudnikOleksii (approved):
👍

PR #1475, opened 2023-11-30
Solution

BoblykNazar:
[DEMO LINK](https://BoblykNazar.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/interfaces/FilterParams.ts
+++ b/src/interfaces/FilterParams.ts
@@ -1,0 +1,4 @@
+export interface FilterParams {
+  sortField: string,
+  reverse: boolean
+}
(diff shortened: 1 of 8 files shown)


src/interfaces/SortType.ts
```diff
@@ -0,0 +1,4 @@
+export enum SortType {
+  name = 'name',
+  length = 'length',
+}
```
emilionila, line 4: you should add default value into SortType

`Use enum SortType instead of string literals for possible sorting options (including the default one)`

src/interfaces/FilterParams.ts
```diff
@@ -0,0 +1,4 @@
+export interface FilterParams {
+  sortField: string,
```
BudnikOleksii, line 2: `sortField` isn't string, it should be `SortType`

BudnikOleksii (changes requested):
Good job 👍 
Just a few moments with `sortField` type should be fixed

PR #1476, opened 2023-11-30
finish task

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer : string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer : string[] = [
```
Anton-Kuchmasov, line 7: you could create an own type
etc. 

`type GoodsList = string[]`

btw, in that case it's not mandatory =)

src/App.tsx
```diff
@@ -15,48 +17,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
```
Anton-Kuchmasov, line 20: good!

src/App.tsx
```diff
@@ -15,48 +17,110 @@ export const goodsFromServer = [
...
+  if (reversed) {
+    return preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortType.Default);
+  const [reversed, setReversed] = useState(false);
```
Anton-Kuchmasov, line 59: AFAIK, it's better to use Generics via using `useState`

like that:

`const [sortField, setSortField] = useState<string>('');`

src/App.tsx
```diff
@@ -15,48 +17,110 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map(good => (
+          <li
+            data-cy="Good"
+            key={uuidv4()}
```
Anton-Kuchmasov, line 119: no idea how it works =))

Anton-Kuchmasov (approved):
nice job!

in future, try use a <Generics> into hook's declaration

src/App.tsx
yefimov77: > no idea how it works =))

it gives random number to a key.property, just like id)

src/App.tsx
```diff
@@ -15,48 +17,110 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  Default = '',
+  Name = 'alphabeticaly',
+  Length = 'length',
+}
+
+interface FunctionArguments {
+  sortField : string,
```
BudnikOleksii, line 27: `sortField` should have `SortType` type instead of string

src/App.tsx
BudnikOleksii: > AFAIK, it's better to use Generics via using `useState`
> 
> like that:
> 
> `const [sortField, setSortField] = useState<string>('');`

When you set default value to `useState` it understand what type it should, so you don't need to specify with generic in this case

src/App.tsx
```diff
@@ -15,48 +17,110 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortField || reversed) && (
```
BudnikOleksii, line 104: Create a variable for `sortField || reversed` condition to avoid mixing different logical operators

src/App.tsx
BudnikOleksii: > > no idea how it works =))
> 
> it gives random number to a key.property, just like id)

It's a bad idea to set the key like this, it's the same as if you use index. As on every render `uuid()` call generates new keys.
You should prepare keys before or just use `good`, in this it's enough

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them

PR #1480, opened 2023-11-30
solution

mickhakk:


src/App.tsx
```diff
@@ -15,47 +16,104 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  none = '',
+  alphabetically = 'alphabetically',
+  length = 'length',
```
ArtemTeslenko, line 22:
```suggestion
  None = '',
  Alphabetically = 'alphabetically',
  Length = 'length',
```

src/App.tsx
```diff
@@ -15,47 +16,104 @@ export const goodsFromServer = [
...
+  sortMethod: SortType;
+  isReversed: boolean;
+}
+
 export const App: React.FC = () => {
+  const getGoodsToView = (
+    goods: string[],
+    { sortMethod, isReversed }: SortingOptions,
+  ) => {
+    let newGoods = [...goods];
```
ArtemTeslenko, line 35: why `let`? you will have ability to change array if it declared as `const`

src/App.tsx
```diff
@@ -15,47 +16,104 @@ export const goodsFromServer = [
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortMethod || isReversed) && (
```
ArtemTeslenko, line 99: move long conditions into variable

ArtemTeslenko (approved):
Nice! I am approving, but see comments 😉

PR #1498, opened 2023-12-05
solution

Barosz30:
https://barosz30.github.io/react_list-of-goods-hooks/


src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+  const goods = getReorderedGoods(
+    [...goodsFromServer], { sortType, isReversed },
+  );
+
+  const sort = (text: SortType) => {
+    setSortType(text);
+  };
```
kate-prakofyeva, line 68: Here you can rewrite as Paul showed with carrying:
```
const handleSort = (type: SortType) => () => {
    setSortType(type);
  };
```
and use it in this way onClick={handleSort(SortType.ALPHABET)}, without callback

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed)
+        && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => reset(SortType.NONE)}
```
kate-prakofyeva, line 117: There is no need to pass an argument here `onClick={() => reset(SortType.NONE)}`, since we are setting default values in the function, like with `setIsReversed(false)`
const reset = (text: SortType) => {
    setSortType(text); <------------------SortType.NONE
    setIsReversed(false);
  };

kate-prakofyeva (approved):
Good job! 
I don’t know why I get your task for review every time :)

src/App.tsx
```diff
@@ -15,47 +16,116 @@ export const goodsFromServer = [
...
+  }
+
+  return visibleGoods;
+};
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
+  const goods = getReorderedGoods(
+    [...goodsFromServer], { sortType, isReversed },
```
JoyCoffeeAddict, line 63: you spread this argument in a function itself, so it's not needed here again

JoyCoffeeAddict (approved):

PR #1505, opened 2023-12-07
Solution

zaura333:
[DEMO LINK](https://zaura333.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType,
+  isReversed: boolean,
+};
+
+export function getReorderedGoods(
```
jacekkosiba, line 30: types could be moved to a separate files
original declaration of
`export function getReorderedGoods(`
seems tu suggest that it should

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
+  goods: string[],
+  { sortType, isReversed }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  if (sortType === SortType.ALPHABET) {
```
jacekkosiba, line 36: This is 100% fine, although I would personally use switch - it works great with enum
`switch (sortType) {
    case NONE:
      break;
    case ALPHABET:
      visibleGoods.sort...` etc

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button', 'is-info', {
+            'is-light': sortType !== SortType.ALPHABET,
+          })}
+          onClick={handleSort(SortType.ALPHABET)}
```
mvjl000, line 78: This `handleSort()` gets invoked once the JSX is being rendered. To avoid it you should either declare a handling function and pass it to onClick e.g. `onClick={handleSomeClick}` or do it this way:

```suggestion
          onClick={() => handleSort(SortType.ALPHABET)}
```

By doing so, the function is invoked ON CLICK as it should be ;)

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={classNames('button', 'is-success', {
+            'is-light': sortType !== SortType.LENGTH,
+          })}
+          onClick={handleSort(SortType.LENGTH)}
```
mvjl000, line 88: Same here

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={classNames('button', 'is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={handleReverse}
```
mvjl000, line 98: Here it's ok, because you are not invoking it ;)

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, { sortType, isReversed })
+            .map((good) => {
```
mvjl000, line 119: I personally prefer, when the JSX is as clean as possible. I would assign it to a constant above the return:
```
const goodsToRender = getReorderedGoods(goodsFromServer, { sortType, isReversed });

// and inside the jsx:

{goodsToRender.map(stays the same)}
```

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
+  } else if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
mvjl000, line 47: Remove all unnecessary comments so your code is 'merge-ready' ;)

src/App.tsx
mvjl000: Agree, in this case it would look cleaner, more structured. Simply more readable ;)

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, { sortType, isReversed })
+            .map((good) => {
+              return (
+                <li data-cy="Good" key={good}>{good}</li>
+              );
+            })}
```
mvjl000, line 123: You could use a shorthand
```suggestion
            .map((good) => (
                <li data-cy="Good" key={good}>{good}</li>
              ))}
```

src/App.tsx
```diff
@@ -15,47 +16,111 @@ export const goodsFromServer = [
...
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+
+  const [sortType, setSortType] = useState(SortType.NONE);
```
mvjl000, line 55: When I work with state & enums I like to add a type for it
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.NONE);
```

PR #1506, opened 2023-12-07
solution

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,103 @@
-import React from 'react';
+import { useState, FC } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { getReorderedGoods } from './functions';
+import { SortType } from './types';
+import goodsFromServer from './goodsFromServer';
 
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+export const App: FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
 
-export const App: React.FC = () => {
+  const sortAlphabetically = () => {
+    if (sortType !== SortType.ALPHABET) {
+      setSortType(SortType.ALPHABET);
+    }
+  };
+
+  const sortByLength = () => {
+    if (sortType !== SortType.LENGTH) {
+      setSortType(SortType.LENGTH);
+    }
+  };
+
+  const reverseList = () => {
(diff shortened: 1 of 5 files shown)


MarcinLigmann (approved):
looks perfect👍

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {
+            visibleGoods.map(item => {
+              return (
+                <li key={item} data-cy="Good">{item}</li>
+              );
+            })
+          }
```
mvjl000, line 100: I would go for a shorthand
```suggestion
          {
            visibleGoods.map(item => (
                <li key={item} data-cy="Good">{item}</li>
              ))
          }
```

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+          onClick={sortByLength}
         >
           Sort by length
         </button>
-
         <button
           type="button"
-          className="button is-warning is-light"
+          className={reverseBtnClass}
+          // className="button is-warning is-light"
```
mvjl000, line 74: Remove unnecessary comments ;)

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+    setSortType(SortType.NONE);
+  };
+
+  const visibleGoods = getReorderedGoods(goodsFromServer, {
+    isReversed,
+    sortType,
+  });
+
+  const alphabetIsActive = sortType === SortType.ALPHABET;
+  const lengthIsActive = sortType === SortType.LENGTH;
```
mvjl000, line 40: Since you are using both of these in only one place, I don't think it's necessary to create a constant for them

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+    if (sortType !== SortType.ALPHABET) {
+      setSortType(SortType.ALPHABET);
+    }
+  };
+
+  const sortByLength = () => {
+    if (sortType !== SortType.LENGTH) {
+      setSortType(SortType.LENGTH);
+    }
+  };
```
mvjl000, line 23: Cool that, you have thought about those if statements (not necessary in this case but still cool)

You can leave it that way, but you could also go for a even better solution (hoping that it won't make the tests mad 😆)! Get rid off those if statements and disable buttons, basing on state:

`disabled={sortType === SortType.ALPHABET}`

src/App.tsx
```diff
@@ -1,61 +1,103 @@
...
+
+  const reverseList = () => {
+    setIsReversed(prevIsReversed => !prevIsReversed);
```
mvjl000, line 26: In this case, I would simply do:
```suggestion
    setIsReversed(!isReversed);
```

PR #1507, opened 2023-12-07
Solution

MarcinLigmann:
[DEMO LINK](https://MarcinLigmann.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,18 +24,16 @@ enum SortType {
 
 type ReorderOptions = {
   sortType: SortType,
-  reverse: boolean,
+  isReverse: boolean,
 };
 
 export function getReorderedGoods(
   goods: string[],
-  { sortType, reverse }: ReorderOptions,
+  { sortType, isReverse }: ReorderOptions,
 ) {
   const visibleGoods = [...goods];
 
   switch (sortType) {
-    case SortType.NONE:
-      break;
     case SortType.ALPHABET:
       visibleGoods.sort((a, b) => a.localeCompare(b));
       break;
@@ -46,7 +44,7 @@ export function getReorderedGoods(
       break;
   }
 
-  if (reverse) {
+  if (isReverse) {
     visibleGoods.reverse();
   }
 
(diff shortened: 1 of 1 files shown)


WiktoriaBielec (approved):
looks fine

src/App.tsx
```diff
@@ -15,47 +16,98 @@ export const goodsFromServer = [
...
+
+export function getReorderedGoods(
+  goods: string[],
+  { sortType, reverse }: ReorderOptions,
+) {
+  const visibleGoods = [...goods];
+
+  switch (sortType) {
+    case SortType.NONE:
+      break;
```
mvjl000, line 38: No need for this case, could just go for default ;)

src/App.tsx
```diff
@@ -15,47 +16,98 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={classNames('button is-info',
+            { 'is-light': sortType !== SortType.ALPHABET })}
+          onClick={() => setSortType(SortType.ALPHABET)}
```
mvjl000, line 75: Create above an event handling function for those onClicks

src/App.tsx
```diff
@@ -15,47 +16,98 @@ export const goodsFromServer = [
...
+  if (reverse) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState(SortType.NONE);
+  const [reverse, setReverse] = useState(false);
```
mvjl000, line 58: A typical convention for boolean state is to prefix them with e.g. `is`. So it would be like `isReverse, setIsReverse`

src/App.tsx
mvjl000:
```
const handleSortAlphabeticaly = () => {
 setSortType(SortType.ALPHABET)
}
// and then
onClick={handleSortAlphabeticaly}

// Or even better one
  const handleSortBy = (value: SortType) => () => {
    setSortType(value);
  };
  // then
  onClick={handleSortBy(SortType.ALPHABET)}
```

src/App.tsx
```diff
@@ -15,47 +16,98 @@ export const goodsFromServer = [
...
+
+  const reset = () => {
+    setSortType(SortType.NONE);
+    setReverse(false);
+  };
+
+  const goods = getReorderedGoods(goodsFromServer, { sortType, reverse });
+  const initialState = (sortType !== SortType.NONE || reverse);
```
mvjl000, line 66: This is quite the opposite of initial state 😆 
I would name it `shouldShowReset`

mvjl000 (changes requested):
Well done, I just have some tweaks for you

PR #1514, opened 2023-12-08
Solution

mosowiecka:
[DEMO LINK](https://mosowiecka.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+// import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
+      case SortType.NONE:
+        return sign * (goodsFromServer.indexOf(good1)
+                  - goodsFromServer.indexOf(good2));
+      default:
+        return 0;
+    }
+  });
+
+  return visibleGoods;
+}
```
mvjl000, line 53: Try to do it the same way as in the previous react_list-of-goods (without the `sign`)

src/App.tsx
```diff
@@ -15,47 +16,118 @@ export const goodsFromServer = [
...
+
+export const App = () => {
+  const [state, setState] = useState({
+    sortType: SortType.NONE,
+    isReversed: false,
+  });
```
mvjl000, line 59: You should split to to two separate states, `sortType` and `isReversed`

mvjl000 (changes requested):

mvjl000 (approved):
Good!

PR #1519, opened 2023-12-11
solution

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,47 +15,130 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={sortType === 'alphabetical'
+            ? 'button is-info'
+            : 'button is-info is-light'}
```
mvjl000, line 98: - Use the enum value

```suggestion
          className={`button is-info ${sortType === SortType.ALPHABETICAL ? '' : 'is-light'}`}
```

When you rename some className, you would have to change it in only one place

src/App.tsx
```diff
@@ -15,47 +15,130 @@ export const goodsFromServer = [
...
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+        {sortType !== '' || isReversed
```
mvjl000, line 124: enum
```suggestion
        {sortType !== SortType.NONE || isReversed
```

src/App.tsx
```diff
@@ -15,47 +15,130 @@ export const goodsFromServer = [
...
+type SortCallback = (a: string, b: string) => number;
+
+enum SortType {
+  LENGTH = 'length',
+  ALPHABETICAL = 'alphabetical',
+  NONE = '',
+}
+
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState<string[]>(goodsFromServer);
```
mvjl000, line 27: In this task, you are not supposed to keep goods in state

mvjl000: Take the same approach as in the previous task - list-of-goods (that one without hooks)

src/App.tsx
```diff
@@ -15,47 +15,130 @@ export const goodsFromServer = [
...
+      }
+
+      return 0;
+    });
+
+    setSortType(SortType.ALPHABETICAL);
+  };
+
+  const sortByLength = () => {
+    let copyGoods = [...goods];
```
mvjl000, line 57: Inside of this compoment, you just need like 3 handler functions, for changing sortType, changing isReverse and reseting the state.

Do all the sorting in a function outside of the component. The function should accept goods from server, sortType and isReverse. Then you will be able to sort items basing on that (ideally using a switch statement).
Return the goods in proper order and render them inside of jsx

mvjl000 (changes requested):

mbruhler (changes requested):
Introduce previouse change requests

mpiskadlo (approved):
Looks good now!

PR #1520, opened 2023-12-12
Solution

Sibunas:
[DEMO LINK](https://sibunas.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,49 +15,131 @@ export const goodsFromServer = [
...
+// type State = {
+//   isReversed: boolean,
+//   sortType: SortType,
+// };
+
+export const App = () => {
+  const [reorderOptions, setReorderOptions] = useState<ReorderOptions>({
+    sortType: SortType.NONE,
+    isReversed: false,
+  });
```
JoyCoffeeAddict, line 64: I think it would be much easier and cleaner to just create two separate states: 
```
[sortType, setSortType] ....
[isReserved, setIsReserved]...
```

src/App.tsx
```diff
@@ -15,49 +15,131 @@ export const goodsFromServer = [
...
+    setReorderOptions({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  const visibleGoods = getReorderedGoods(goodsFromServer, reorderOptions);
+  const alphabetButtonClass
+  = reorderOptions.sortType
+  === SortType.ALPHABET ? 'button is-info' : 'button is-info is-light';
```
JoyCoffeeAddict, line 97: This can be set in-line in the JSX itself, because the only thing that you are changing is `is-light` part

JoyCoffeeAddict (changes requested):

src/App.tsx
Sibunas: changes implemented

adrian200002 (approved):

src/App.tsx
```diff
@@ -15,49 +15,118 @@ export const goodsFromServer = [
...
+  const handleReverse = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const handleReset = () => {
+    setSortType(SortType.NONE);
+    setIsReversed(false);
+  };
+
+  const visibleGoods
```
mbruhler, line 81: This is bad formatting of code, please make it so it looks proper. You can break on ? or : but not on === or =

Applies everywhere

src/App.tsx
```diff
@@ -15,49 +15,118 @@ export const goodsFromServer = [
...
+  === SortType.ALPHABET ? '' : 'is-light';
+  const reverseButtonClass
+  = isReversed
+    ? '' : 'is-light';
+  const lengthButtonClass
+  = sortType
+  === SortType.LENGTH ? '' : 'is-light';
+
+  return (
+    <div>
```
mbruhler, line 94: Missing orginal classes and html markup propably, so it's looking bad

mbruhler (changes requested):
In case of working, it works. Still, doesn't look because of wrong html markup and is badly formatted. Please fix this in order to approve

PR #1550, opened 2024-01-05
react_list-of-goods-hooks

serjgrm:
e [DEMO LINK](https://serjgrm.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,105 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+const adoptedGoods = goodsFromServer.map((good, id) => ({ name: good, id }));
```
denys-danyliuk, line 19:
```suggestion
const adoptedGoods = goodsFromServer.map(({ good, id }) => ({ name: good, id }));
```
your code will not took the id from the good, it will use index as it is second param in map callback function

src/App.tsx
```diff
@@ -15,48 +16,105 @@ export const goodsFromServer = [
...
+
+enum SortType {
+  ALPHABETIC = 'ALPHABETIC',
+  LENGTH = 'LENGTH',
+}
+
+export const App = () => {
+  const [goods, setState] = useState(adoptedGoods);
+  const [sortedByState, setsortedByState]
```
denys-danyliuk, line 28:
```suggestion
  const [sortedByState, setSortedByState]
```

denys-danyliuk: also, do you need separate value for this in state?

src/App.tsx
```diff
@@ -15,48 +16,105 @@ export const goodsFromServer = [
...
+    if (value === SortType.ALPHABETIC) {
+      modifiedGoods.sort((good1, good2) => (
+        good1.name.localeCompare(good2.name)));
+      setsortedByState(SortType.ALPHABETIC);
+      setState(modifiedGoods);
+
+      return;
+    }
+
+    if (value === SortType.LENGTH) {
```
denys-danyliuk, line 44: use switch statement

denys-danyliuk (changes requested):

lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #1590, opened 2024-01-18
add task solution

yepolotn1ak:
[DEMO LINK](https://yepolotn1ak.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,48 +15,123 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={`
+            button
+            is-warning
+            ${isReversed ? '' : 'is-light'}
```
etojeDenys, line 108: better to use classnames package

src/App.tsx
```diff
@@ -15,48 +15,123 @@ export const goodsFromServer = [
...
 
+interface FilterParams {
+  sortField: string;
+  isReversed: boolean;
+}
+
+type IsLightFunction = (a: string, b: string) => string;
+
+const SORT_FILED_ALPHABET = 'Sort alphabetically';
+const SORT_FILED_LENGTH = 'Sort by length';
```
etojeDenys, line 26: better to create enum instead of several constants

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,125 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum Sortfield {
+  SORT_FILED_ALPHABET = 'Sort alphabetically',
+  SORT_FILED_LENGTH = 'Sort by length',
+}
+
+interface FilterParams {
+  sortField: Sortfield | string;
```
anastasiia-tilikina, line 25: this type doesn't restrict your sort field value since I can put here any string. it is better to add to enum value for default sorting.

src/App.tsx
```diff
@@ -15,48 +16,125 @@ export const goodsFromServer = [
...
+  }
+
+  if (isReversed) {
+    preparedGoods = preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
+const isLightStyle: IsLightFunction = (sortField, sortFieldType) => {
```
anastasiia-tilikina, line 57: with classnames library you don't need this function. there is a way to add conditional classes provided by this library itself. you jus need to add object where is properly is class and value is condition when this class should be applied.

src/App.tsx
```diff
@@ -15,48 +16,125 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={cn(`
+            button
+            is-info
+            ${isLightStyle(sortField, Sortfield.SORT_FILED_ALPHABET)}
```
anastasiia-tilikina, line 87: like this: 
```suggestion
           {'is-light': sortField === sortFieldType}
```

src/App.tsx
```diff
@@ -15,48 +16,125 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={cn(`
+            button
+            is-success
+            ${isLightStyle(sortField, Sortfield.SORT_FILED_LENGTH)}
```
anastasiia-tilikina, line 99: fix everywhere

src/App.tsx
anastasiia-tilikina: and here not fixed as well

anastasiia-tilikina (changes requested):
GJ

check my comments

etojeDenys (approved):
lgtm

PR #1603, opened 2024-01-27
Develop

Diliradon:
https://Diliradon.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,135 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+
+enum SortType {
+  SORT_ALPHABETICALLY = 'Sort alphabetically',
+  SORT_BY_LENGTH = 'Sort by length',
+  DEFAULT = '',
+}
 
 export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+  const preparedGoods = [...goods];
+
+  if (sortFild) {
+    preparedGoods.sort(({ good: good1 }, { good: good2 }) => {
+      switch (sortFild) {
+        case SortType.SORT_ALPHABETICALLY:
+          return good1.localeCompare(good2);
+        case SortType.SORT_BY_LENGTH:
+          return good1.length - good2.length;
+        default:
```
l4st1m0za, line 48: The default case in the sort switch statement returns 0. It would be better to return the original array in the default case instead of 0.

src/App.tsx
```diff
@@ -1,61 +1,135 @@
...
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortFild, setSortFild] = useState(SortType.DEFAULT);
+  const [reverseFild, setReverseFild] = useState(false);
+
+  const visibleGoods = getPreparedGoods(
+    goodsFromServer,
```
l4st1m0za, line 66: The getPreparedGoods function is called inside the component rendering. Keep in mind that this function will be called on every render of the component. It would be better to use a useEffect hook to handle sorting and reversing.

l4st1m0za (approved):
Great job, consider the comments for the future tasks

PR #1606, opened 2024-01-28
solution

EvgenGolub:
[DEMO LINK](https://evgengolub.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,48 +16,86 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const SORT_FIELD_ALPH = 'Sort alphabetically';
+const SORT_FIELD_LENGTH = 'Sort by length';
```
etojeDenys, line 20: you can create the enum instead of const vars

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,48 +16,88 @@ export const goodsFromServer = [
...
 ];
 
+enum Sort {
+  alph = 'Sort alphabetically',
+  length = 'Sort by length',
+}
+
+function getPreparedGoods(
+  goods: string[],
+  sortField: string,
```
```suggestion
  sortField: Sort,
```

src/App.tsx
```diff
@@ -15,48 +16,88 @@ export const goodsFromServer = [
...
+          return 0;
+      }
+    });
+  }
+
+  return isReversed ? preparedGoods.reverse() : preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
```
etojeDenys, line 50:
```suggestion
  const [sortField, setSortField] = useState(Sort.Default);
```
add it to the enum too

etojeDenys (changes requested):

src/App.tsx
EvgenGolub: My bad

DarkMistyRoom (approved):
Well done!

PR #1635, opened 2024-02-15
Solution

Krzychu92:
https://krzychu92.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,61 +1,116 @@
-import React from 'react';
-import 'bulma/css/bulma.css';
-import './App.scss';
+import React, { useEffect, useState } from "react";
+import "bulma/css/bulma.css";
+import "./App.scss";
 
 export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  "Dumplings",
+  "Carrot",
+  "Eggs",
+  "Ice cream",
+  "Apple",
+  "Bread",
+  "Fish",
+  "Honey",
+  "Jam",
+  "Garlic",
 ];
 
+export function getReorderedGoods(
+  goods: string[],
+  sortType: string,
+  isReversed: boolean,
+) {
+  // To avoid the original array mutation
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((good1: string, good2: string): number => {
+    if (sortType === "ALPHABET") {
+      return good1.localeCompare(good2);
+    }
+
+    if (sortType === "LENGTH") {
+      return good1.length - good2.length;
+    }
+
+    if (sortType === "NONE") {
+      return 0;
+    }
+
+    return 0;
+  });
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState("NONE");
+  const [isReversed, setIsReversed] = useState(false);
+  const goodies = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
+  useEffect(() => {}, [isReversed]);
+
   return (
     <div className="section content">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+  "Jam",
+  "Garlic",
 ];
 
+export function getReorderedGoods(
+  goods: string[],
+  sortType: string,
+  isReversed: boolean,
+) {
+  // To avoid the original array mutation
```
BogdanMaliuta, line 23: delete comments

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+
+    if (sortType === "LENGTH") {
+      return good1.length - good2.length;
+    }
+
+    if (sortType === "NONE") {
+      return 0;
+    }
+
+    return 0;
```
BogdanMaliuta, line 39: same return 0, you can reove your condition

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState("NONE");
+  const [isReversed, setIsReversed] = useState(false);
+  const goodies = getReorderedGoods(goodsFromServer, sortType, isReversed);
+
+  useEffect(() => {}, [isReversed]);
```
BogdanMaliuta, line 53: what does this useEffect?

src/App.tsx
```diff
@@ -1,61 +1,116 @@
...
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={
+            sortType !== "ALPHABET"
+              ? "button is-info is-light"
+              : "button is-info"
+          }
```
BogdanMaliuta, line 64: use ClassNames library for better styles conditions

BogdanMaliuta (changes requested):
Not bad, fix some mistakes, and I will aprove it

src/App.tsx
Krzychu92: how about now?

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+export function getReorderedGoods(
+  goods: string[],
+  sortType: string,
+  isReversed: boolean,
+) {
+  const visibleGoods = [...goods];
+
+  visibleGoods.sort((itemOne: string, itemTwo: string): number => {
+    switch (sortType) {
+      case "ALPHABET":
```
yevhenii-pyl, line 28: Let's use enum for it. As a rule of thumb, it's recommended to avoid using plain strings for switch cases.

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+  });
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState("NONE");
```
yevhenii-pyl, line 44: Can't see 'NONE' as a case at your switch statement. 
In this case, it's better to pass `undefined` as a initial value if the idea is to trigger default option at switch.

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
           Reverse
         </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {sortType !== "NONE" || isReversed ? (
```
yevhenii-pyl, line 80: Having `undefined` instead of "NONE" may simplify this condition to `sortType || isReversed`

src/App.tsx
```diff
@@ -1,61 +1,105 @@
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
```
yevhenii-pyl, line 84: It's better to extract this logic into a separate handler. Who knows, it can be extended at future.

yevhenii-pyl (changes requested):
Way better! 
Let's address a few minor things and the job is done 👍

yevhenii-pyl (approved):
Good job 🚀

PR #1641, opened 2024-02-16
Added task solution

OleksiiNesteruk:
- [DEMO LINK](https://OleksiiNesteruk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import cn from 'classnames';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,36 +16,100 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
+  ALPH = 'alphabetically',
+  LENGTH = 'length',
+  NONE = '',
+}
```
volodymyr-soltys97, line 23: Consider will be better if you move enum to the separate file

volodymyr-soltys97 (approved):
Great work 🔥

PR #1655, opened 2024-02-25
solution

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


yevhenii-pyl:
Tests are failing:

```
      1) should allow to reset

    after double reverse



  12 passing (10s)
  1 failing


  1) Page
       after pressing "Reverse"
         should allow to reset:

      Timed out retrying after 4000ms
      + expected - actual

      -'Garlic'
      +'Dumplings'
      
      at Object.assertFirstGoods (http://localhost:3001/__cypress/tests?p=cypress/integration/page.spec.js:106:24)
      at Context.eval (http://localhost:3001/__cypress/tests?p=cypress/integration/page.spec.js:173:12)

```

src/App.tsx
```diff
@@ -15,35 +16,120 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${cn({
+            'is-light': sortType !== SortType.ALPHABET,
+          })}`}
+          onClick={() => {
```
yevhenii-pyl, line 78: Not cool :) 

You already have a `sortByAplhabet ` handler, why don't you put all logic in it? 
Basically, that's the idea of handlers.

src/App.tsx
```diff
@@ -15,35 +16,120 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={`button is-success ${cn({
+            'is-light': sortType !== SortType.LENGTH,
+          })}`}
+          onClick={() => {
+            sortByLength();
```
yevhenii-pyl, line 92: Same here, by using handlers you really want to incapsulate some logic inside.

src/App.tsx
```diff
@@ -15,35 +16,120 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-warning ${cn({
+            'is-light': isReversed !== true,
+          })}`}
+          onClick={() => {
+            setIsReversed(!isReversed);
```
yevhenii-pyl, line 105: Same with handler here.
Also, when you want to use previous state to manipulate next state with useState hook it's always better to do it that way: 
`setState(prevState => !prevState)`

yevhenii-pyl: Also, somewhere here you have that unwanted behaviour hidden that leads to the failed test.

yevhenii-pyl (changes requested):
Good job! Let's adjust your code a bit and it's good to go.

yevhenii-pyl (approved):
Good job 👍

PR #1658, opened 2024-02-26
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_list-of-goods-hooks/)

I decided not to use the function for sorting. Wanted to try a different approach

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,37 +16,104 @@ export const goodsFromServer = [
...
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn({
+            'button is-warning': true,
+            'is-light': !isReversed,
+          })}
+          onClick={() => {
+            setIsReversed(!isReversed);
+            setGoods(goods.reverse());
```
nazarmatsevych, line 85: Better to create a separate function and use it here to make JSX cleaner

src/App.tsx
```diff
@@ -15,37 +16,104 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortType !== null || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setGoods(goodsFromServer);
+              setSortType(null);
+              setIsReversed(false);
```
nazarmatsevych, line 98: same here

nazarmatsevych (changes requested):
Almost done, please review the comments below

l4st1m0za (approved):
Good job!

PR #1659, opened 2024-02-26
added solution react list of goods hooks

nk8bnj:
[DEMO LINK](https://nk8bnj.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


HolyRaveOS (approved):
Good joob!

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
...
+          className={`button is-info ${sortMethod !== SortMethod.Alphabet && 'is-light'}`}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          onClick={() => setSortMethod(SortMethod.Length)}
+          type="button"
+          className={`button is-success ${sortMethod !== SortMethod.Length && 'is-light'}`}
```
BudnikOleksii, line 78: Use classnames library for conditional setting classes https://www.npmjs.com/package/classnames. As in your case you set classes `false` if a condition is false
image

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
...
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          onClick={() => setIsListReversed(!isListReversed)}
```
BudnikOleksii, line 84:
```suggestion
          onClick={() => setIsListReversed(prev => !prev)}
```
If you need to set a state depending on the previous one use functional setState.

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

BudnikOleksii (approved):
Looks great 🚀

PR #1681, opened 2024-03-07
works but fail test on reset visibility

spencerspe:
[DEMO LINK](https://spencerspe.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,34 +16,103 @@ export const goodsFromServer = [
...
+          }}
+          className='button is-danger is-light'
+          onClick={reset}
+        >
           Reset
         </button>
       </div>
 
       <ul>
         <ul>
```
lerastarynets, line 110: you need only 1 ul tag

lerastarynets (changes requested):
Unfortunately, your project is not reachable via the demo link(

PR #1682, opened 2024-03-07
add solution

RusLan2186:
DEMO LINK](https://RusLan2186.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +16,106 @@ export const goodsFromServer = [
...
+  sortField: SortType;
+}
+
+function getPreparedGoods(
+  goods: string[],
+  { sortField, isReversed }: FilterParams,
+) {
+  const preparedGoods = [...goods];
+
+  if (sortField) {
```
Vasya564, line 36: There is no need to check sortField. If we don't pass sortField the ```default``` case will work

Vasya564 (approved):
Good job!🔥 Check my comment

src/App.tsx
BudnikOleksii: > There is no need to check sortField. If we don't pass sortField the `default` case will work

But you actually don't need to sort if it's not passed, so the check is a good idea

BudnikOleksii (approved):
Looks great 🚀

PR #1708, opened 2024-03-14
add solution

tetlisna:
[DEMO LINK](https://tetlisna.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+          className={cn('button is-warning', { 'is-light': !isReverse })}
+          onClick={() => setIsReversed(!isReverse)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortField || isReverse) && (
```
ivan-gritsenko, line 95: Create variable for `sortField || isReverse`

ivan-gritsenko (approved):
LGTM!🔥

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  alphabetical = 'alphabetical',
+  length = 'length',
+}
+
+function getPrepareGoods(
+  goods: string[],
+  field: string,
+  reversed: boolean,
```
BudnikOleksii, line 28: - What field?
- Boolean variables/arguments should start with `is` or `has`

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+): string[] {
+  let prepareGoods = [...goods];
+
+  if (field === SortType.alphabetical) {
+    prepareGoods.sort((a: string, b: string) => a.localeCompare(b));
+  } else if (field === SortType.length) {
```
BudnikOleksii, line 34: Use switch case instead

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+  }
+
+  return prepareGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [isReverse, setIsReversed] = useState(false);
+
+  const updatedGoods = getPrepareGoods(goodsFromServer, sortField, isReverse);
```
BudnikOleksii, line 49:
```suggestion
  const preparedGoods = getPrepareGoods(goodsFromServer, sortField, isReverse);
```

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+
+  const handleclick = (field: string) => {
+    if (field === sortField) {
+      setIsReversed(!isReverse);
+      setSortField(field);
+    }
```
BudnikOleksii, line 55: - When I click on any sort type button it toggles `reverse`, and it doesn't make sense 🤷. The Sort button should only set the corresponding sort type, reverse button toggle the `isReversed` state.
- The name is not good too, what click does it handle?
- When you need to change state depending on the previous one use functional `setState`. `setIsReversed(prev => !prev)`

src/App.tsx
BudnikOleksii: > Create variable for `sortField || isReverse`

Good point

BudnikOleksii (changes requested):
Good start 👍 
Left some comments, check them

StasSokolov1 (approved):
Great job🔥

PR #1711, opened 2024-03-14
task solution

Math8585:
- [DEMO LINK](https://Math8585.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
```
vladarskyi, line 19: As I understand it, in `README` they are asking to add a default sorting option to enam:

`Use enum SortType instead of string literals for possible sorting options (including the default one)`

vladarskyi (changes requested):
Good job! one moment from README and I think it would be good to add a type to the array with products.

src/App.tsx
```diff
@@ -15,37 +16,231 @@ export const goodsFromServer = [
...
+    preperedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.Alphabet:
+          return good1.localeCompare(good2);
+
+        case SortType.Length:
+          return good1.length - good2.length;
+
+        default:
+          return 0;
```
polosanya, line 43: Would be great to move switch...case out of the sort callback to avoid calling sort method if the case is default

src/App.tsx
```diff
@@ -15,37 +16,231 @@ export const goodsFromServer = [
...
+
+  return preperedGoods;
+}
+
+export const App = () => {
+  const [sortField, setsortField] = useState<SortType>(SortType.Default);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const visibleGoods = getPreperedGoods({ sortField, isReversed });
+  const sortedOrReversedList = sortField || isReversed;
```
polosanya, line 60:
```suggestion
  const isResetActive= sortField || isReversed;
```
Let's simplify this variable

src/App.tsx
```diff
@@ -15,37 +16,231 @@ export const goodsFromServer = [
...
+
+  const resetSortHandler = (): void => {
```
polosanya, line 62:
```suggestion
  const reset= (): void => {
```

src/App.tsx
```diff
@@ -15,37 +16,231 @@ export const goodsFromServer = [
...
+        ))}
       </ul>
     </div>
   );
 };
+
+// import React, { useState } from 'react';
+// import cn from 'classnames';
+// import 'bulma/css/bulma.css';
+// import './App.scss';
```
polosanya, line 125: Don't forget to remove the comments

polosanya (changes requested):

StasSokolov1 (approved):
Great job🔥

PR #1712, opened 2024-03-14
add task solution

vladarskyi:
[DEMO LINK](https://vladarskyi.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +18,110 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setSortBy(SortBy.Name)}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={cn('button', 'is-info', {
```
Vlobdam, line 90: This had is-success class instead of is-info

src/App.tsx
```diff
@@ -15,36 +18,110 @@ export const goodsFromServer = [
...
+
+  if (isReverseAdded) {
+    return sortedGoods.reverse();
+  }
+
+  return sortedGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortBy, setSortBy] = useState<SortBy | null>(null);
```
Vlobdam, line 59: instead of making null default you could use SortBy.Default
```suggestion
  const [sortBy, setSortBy] = useState<SortBy>(SortBy.Default);
```

Vlobdam (approved):
Good but small mistake

src/App.tsx
```diff
@@ -15,36 +18,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortBy {
+  Default,
+  Name,
+  Length,
```
```suggestion
  Default = '',
  Name = 'name',
  Length = 'length',
```

src/App.tsx
BudnikOleksii: > instead of making null default you could use SortBy.Default

Good point

src/App.tsx
BudnikOleksii: > This had is-success class instead of is-info

Good point, the color is different in demo

BudnikOleksii (changes requested):
Good work 👍 
Approved, but check the comments and fix some moments

Moroz-Dmytro (approved):
Looks good 👍

PR #1713, opened 2024-03-15
add solution

Vlobdam:


vlesia (approved):
Good job!

src/App.tsx
```diff
@@ -15,36 +19,108 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = React.useState<SortField>(SortField.All);
+  const [isReverse, setIsReverse] = React.useState(false);
+
+  const sortedGoods: Goods = getSortedGoods(goodsFromServer, {
+    sortField,
+    isReverse,
+  });
+  const isChanged: boolean = isReverse || sortField !== '';
```
BudnikOleksii, line 66:
```suggestion
  const isResetButtonVisible = isReverse || sortField;
```
You can just check if a value exists, and name could be better

src/App.tsx
```diff
@@ -15,36 +19,108 @@ export const goodsFromServer = [
...
+
+  const resetOnClick = () => {
```
BudnikOleksii, line 68:
```suggestion
  const handleReset = () => {
```
In accordance with convention handlers should start from work `handle`
https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

BudnikOleksii (changes requested):
Deploy your solution and add a demo link to the PR so a reviewer can test it

src/App.tsx
```diff
@@ -15,36 +19,108 @@ export const goodsFromServer = [
...
         </button>
-
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {isResetButtonVisible && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => handleReset()}
```
StasSokolov1, line 111:
```suggestion
            onClick={handleReset}
```

can be a bit simplify

StasSokolov1 (approved):
Looks good to me, I'm approving it with only one minor comment🔥

PR #1714, opened 2024-03-15
solution

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/react_list-of-goods-hooks/)


yevhenii-pyl (approved):
Solid: clean and readable! 🚀 

A bit of overkill with sort helpers, but can't call it a bad decision. It's just a matter of preference and the need to consider future improvements/additions in your code that may actually evolve that logic. 
Probably, I would prefer to create a singe helper like `handleSort` that receives a sort type parameter. 
But anyways, great job!

PR #1715, opened 2024-03-15
Add task solution

hattoriKanto:
[DEMO LINK](https://hattoriKanto.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,13 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+import { SortType } from './types/SortType';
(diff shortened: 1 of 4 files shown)


winniepuhh (approved):
Nice!!!!

src/App.tsx
```diff
@@ -15,36 +20,100 @@ export const goodsFromServer = [
...
+
+  if (sortField) {
+    preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.byName:
+          return good1.localeCompare(good2);
+        case SortType.byLength:
+          return good1.length - good2.length;
+        default:
+          return 0;
```
polosanya, line 38: Would be great so move switch...case out of the sort method callback, so you don't need to call it if it's a default case

polosanya (approved):
Good job!

PR #1721, opened 2024-03-16
redo from jsx to tsx

winniepuhh:
[DEMO LINK](https://winniepuhh.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +16,104 @@ export const goodsFromServer = [
...
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((good1, good2) => {
+        if (good1.length === good2.length) {
+          return good1.localeCompare(good2);
+        }
+
+        return good1.length - good2.length;
+      });
+      break;
```
dima8528, line 39: You don't need to duplicate your logic with sort method. You can just take out this method and write switch inside

src/App.tsx
```diff
@@ -15,36 +16,104 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  ALPHABETIC = 'alphabet',
+  LENGTH = 'length',
+}
+
+const getGoods = (goods: string[], sortType: string, isReversed: boolean) => {
```
dima8528, line 24: `sortType` attribute must have the union type `SortType` (that enum) or '' . You should do sorting only if `sortType` is not an empty value. So you need to change your logic a bit

src/App.tsx
```diff
@@ -15,36 +16,104 @@ export const goodsFromServer = [
...
+      visibleGoods.sort((good1, good2) => {
+        if (good1.length === good2.length) {
+          return good1.localeCompare(good2);
+        }
+
+        return good1.length - good2.length;
+      });
+      break;
+    default:
+      break;
```
dima8528, line 41: You don't need to use `break` in the default case. It is already the last case interpreter check

dima8528 (changes requested):
Almost done! Fix these small issues pls and everything will be ok )

src/App.tsx
```diff
@@ -15,36 +17,112 @@ export const goodsFromServer = [
...
+): string[] => {
+  const visibleGoods: string[] = [...goods];
+
+  switch (sortType) {
+    case SortType.ALPHABETIC:
+    case SortType.LENGTH:
+      visibleGoods.sort(sortFunctions[sortType]);
+      break;
+    default:
+  }
```
BudnikOleksii, line 51:
```suggestion
  if (sortType && sortFunctions[sortType]) {
    visibleGoods.sort(sortFunctions[sortType]);
  }
```
- As you created a map with sort functions the switch case is redundant
- `default` case shouldn't be empty

src/App.tsx
```diff
@@ -15,36 +17,112 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  return visibleGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType | ''>('');
```
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.Default);
```

src/App.tsx
```diff
@@ -15,36 +17,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  ALPHABETIC = 'alphabet',
+  LENGTH = 'length',
```
BudnikOleksii, line 22: [CODE STYLE] - Enum constant names usually start with capital letter. Values can have any case that's necessary.
image

BudnikOleksii (changes requested):
Almost done, just a few minor fixes are required

src/App.tsx
winniepuhh: but i do not have Default, even if i will add it in enum SortType it shows me mistake

BudnikOleksii: > but i do not have Default, even if i will add it in enum SortType it shows me mistake

Ok. If you have an error in `sortFunctions` object, you omit this value from your enum with `Exclude` utility type.
https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers
https://stackoverflow.com/questions/48850720/typescript-how-can-omit-some-items-from-an-enum-in-typescript

BudnikOleksii (approved):
👍

PR #1743, opened 2024-03-24
Develop

IhorRome:
[DEMO LINK](https://IhorRome.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,13 +23,13 @@ enum SortType {
 }
 
 type Query = {
-  order: 'asc' | 'desc';
+  reverse: boolean;
 };
 
 function getVisibleGoods(
   goods: Array<string>,
   sortField: SortType,
-  query: Query = { order: 'asc' },
+  query: Query = { reverse: false },
 ) {
   const resGoods = [...goods];
 
@@ -47,7 +47,7 @@ function getVisibleGoods(
     });
   }
 
-  if (query.order === 'desc') {
+  if (query.reverse) {
     resGoods.reverse();
   }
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,35 +16,104 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  Empty = '',
+  Length = 'id',
+  Alphabetically = 'alphabetically',
```
TarasHoliuk, line 22: why `id`? It's better to match value and key if there is no reason to write something else (as with `Empty`)
```suggestion
  Length = 'Length',
  Alphabetically = 'Alphabetically',
```

src/App.tsx
```diff
@@ -15,35 +16,104 @@ export const goodsFromServer = [
...
+  if (query.reverse) {
+    resGoods.reverse();
+  }
+
+  return resGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortType.Empty);
+  const [reverse, setReverse] = useState(false);
```
TarasHoliuk, line 59:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```
for boolean variables: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

TarasHoliuk (changes requested):
GJ :+1: Logic implemented correctly! Added 2 small comments about naming

TarasHoliuk (approved):
Well done :heavy_check_mark:

PR #1754, opened 2024-03-27
solution

robsza1122:
https://robsza1122.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
-
 export const goodsFromServer = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -14,37 +13,117 @@ export const goodsFromServer = [
...
+  return visibleGoods;
+}
+// DON'T save goods to the state
+// type State = {
+//   isReversed: boolean,
+//   sortType: SortType,
+// };
 
 export const App: React.FC = () => {
+  const [order, setOrder] = React.useState<ReorderOptions>({
```
yevhenii-pyl, line 57: I'd suggest having to separate states: for sortType and for isReversed, it's just makes it easier to maintain and debug.

src/App.tsx
```diff
@@ -14,37 +13,117 @@ export const goodsFromServer = [
...
+    isReversed: false,
+  });
+
+  const alphabeticalOrder = () =>
+    setOrder({
+      sortType: SortType.ALPHABET,
+      isReversed: order.isReversed,
+    });
+
+  const orderByLength = () =>
```
yevhenii-pyl, line 68: We can also simplify it here by having a single handler for ordering items, that will receive a sorting type as an argument. If your code and sorting types grows - you may just found yourself flooded with these similar-looking helpers.

src/App.tsx
```diff
@@ -14,37 +13,111 @@ export const goodsFromServer = [
...
+  LENGTH,
+}
+
+type TypeIsReversed = {
+  isReversed: boolean;
+};
+
+type TypeSort = {
+  sortType: number;
+};
```
BogdanMaliuta, line 28: for what you created this types?

src/App.tsx
```diff
@@ -14,37 +13,111 @@ export const goodsFromServer = [
...
+
+export function getReorderedGoods(
+  goods: string[],
+  { isReversed }: TypeIsReversed,
+  { sortType }: TypeSort,
```
BogdanMaliuta, line 33: why your arguments is objects?

src/App.tsx
```diff
@@ -14,37 +13,111 @@ export const goodsFromServer = [
...
+) {
+  const visibleGoods = [...goods];
+
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => (a > b ? 1 : -1));
```
BogdanMaliuta, line 39: using localeCompare method is more safe

BogdanMaliuta (changes requested):
Your buttons should have background colors, check your styles.
Also check my comments

src/App.tsx
robsza1122: localeCompare doesn't work. 

`      visibleGoods.sort((a, b) => a.localeCompare(b));`

src/App.tsx
```diff
@@ -14,37 +14,98 @@ export const goodsFromServer = [
...
+
+  const goods = getReorderedGoods(goodsFromServer, reverse, type);
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is info ${type === SortType.SORT_ALPHABETICALLY ? '' : 'is-light'}`}
```
BogdanMaliuta, line 71: use classNames library for cond classes

BogdanMaliuta (approved):
Hello, looks good.But next time use library classnames for conditional styles

PR #1777, opened 2024-04-03
test

marinatea:
[DEMO LINK](https://marinatea.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +15,127 @@ export const goodsFromServer = [
...
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType;
+  isReversed: boolean;
+};
+
+// Use this function in the render method to prepare goods
```
choeqq, line 29: why do you need this comments?

choeqq (changes requested):

src/App.tsx
marinatea: it is from previous task :) will delete it

choeqq (approved):

PR #1778, opened 2024-04-04
Solution

OlesPartyka:
[DEMO LINK](https://OlesPartyka.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+}
+
+function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+) {
+  const visibleGoods = [...goods];
+
+  switch (true) {
```
choeqq, line 32: if you're using switch why are you switching on `true` instead of `sortType`?

choeqq (changes requested):

src/App.tsx
OlesPartyka: because when I wrote this it worked

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={classNames('button is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={() => {
+            setReverse(prevState => !prevState);
+          }}
```
nataliaklonowska, line 93: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

nataliaklonowska (changes requested):

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.ALPHABET,
+          })}
+          onClick={() => setSortType(SortType.ALPHABET)}
```
nataliaklonowska, line 75: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={classNames('button is-success', {
+            'is-light': sortType !== SortType.LENGTH,
+          })}
+          onClick={() => setSortType(SortType.LENGTH)}
```
nataliaklonowska, line 85: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

nataliaklonowska (changes requested):
remember not only to fix the one I commented on but to find all similar cases in your project

choeqq (approved):

PR #1780, opened 2024-04-04
Solution

pawelkloc:
[DEMO LINK](https://pawelkloc.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


nataliaklonowska (changes requested):
all buttons are highlighted when one of them is active. You should mark only active buttons

src/App.tsx
```diff
@@ -15,36 +15,119 @@ export const goodsFromServer = [
...
+enum SortType {
+  NONE,
+  ALPHABET,
+  LENGTH,
+}
+
+type ReorderOptions = {
+  sortType: SortType;
+  isReversed: boolean;
+  isSorted: boolean;
```
nataliaklonowska, line 27: you never use isSorted so delete it everywhere ;)

nataliaklonowska (changes requested):

src/App.tsx
pawelkloc: Ahh, ok :)

PR #1792, opened 2024-04-10
develop

taras-nakonechnyi:
[DEMO LINK](https://taras-nakonechnyi.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import cn from 'classnames';
+import { Goods } from './Components/Goods/Goods';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/Components/Goods/Goods.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+type Props = {
+  goods: Array<string>;
+};
+
+export const Goods: React.FC<Props> = ({ goods }) => (
+  <ul>
+    <ul>
+      {goods.map((good: string, index: number) => (
+        <li key={index} data-cy="Good">
```
Esceype, line 11:
```suggestion
      {goods.map((good: string) => (
        <li key={good} data-cy="Good">
```

The name will be preferable, since it is unique in this array, it is better not to use the index as a key since the array can change, after which certain operations with the elements of the array may not work correctly, so it is worth using one of the element values, in our case the name of the product itself will be unique.

src/App.tsx
```diff
@@ -15,37 +18,100 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setListReverse(currrent => !currrent)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(!(actualSortMode === SetMode.reset) || listReverse) && (
```
Esceype, line 103: Move this condition to a variable, so it will be easier to read and the JSX part will be less overloaded

Esceype (changes requested):
GJ!
Read the following comments:

nazarmatsevych (approved):
Well done 👍

PR #1802, opened 2024-04-18
commit

rwlp:
[DEMO LINK](https://rwlp.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { useEffect, useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,35 +16,160 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [action, setAction] = useState(SortType.DEFAULT);
+  const [state, setState] = useState(initialState);
+
+  useEffect(() => {
+    if (SortType.NONE !== action) {
+      setState(updateState(state, action));
+    }
+  }, [action]);
```
EdPirro, line 121: `state` must be present in `useEffect`'s dependency array; intentionally omitting a reactive variable in the dependency array is discouraged!

updating your `setState` to `setState(last => updateState(last, action))` should fix this.

Also, your state is behaving much like a reducer; maybe check out [React.useReducer](https://react.dev/reference/react/useReducer) (again, way more complicated than it should be for this simple project)

EdPirro (approved):
Great job, @rwlp! Keep it up! 🚀

src/App.tsx
rwlp: Wow, I really liked these observations

PR #1807, opened 2024-04-22
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_list-of-goods-hooks/)


yevhenii-pyl (changes requested):
Pretty much same comments: let's use the switch statement and create one function for sorting instead of two.

yevhenii-pyl (approved):
Good job 👍 
And same here: let's respect those linter rules :)

PR #1815, opened 2024-04-23
Develop

Reaffith:
[DEMO LINK](https://Reaffith.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,34 +17,92 @@ export const goodsFromServer = [
...
+  }
+
+  const [sortMethod, setSortmethod] = useState<Sort>(Sort.none);
+  const [isReverse, changeReverse] = useState(false);
+  let goods = [...goodsFromServer];
+
+  switch (sortMethod) {
+    case Sort.none:
+      goods = [...goodsFromServer];
+      break;
```
sTorba24, line 33:
```suggestion
    case Sort.none:
      break;
```

sTorba24: it is already with this value

src/App.tsx
```diff
@@ -16,34 +17,92 @@ export const goodsFromServer = [
...
+
+    case Sort.alph:
+      goods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case Sort.length:
+      goods.sort((a, b) => a.length - b.length);
```
sTorba24, line 40: add `default` case to switch

src/App.tsx
```diff
@@ -16,34 +17,92 @@ export const goodsFromServer = [
...
 
 export const App: React.FC = () => {
+  enum Sort {
+    none = '',
+    alph = 'alph',
+    length = 'length',
+  }
+
+  const [sortMethod, setSortmethod] = useState<Sort>(Sort.none);
+  const [isReverse, changeReverse] = useState(false);
```
sTorba24, line 27:
```suggestion
  const [isReversed, changeReversed] = useState(false);
```

sTorba24: reverse is action, state is `reversed` so use `isReversed`

sTorba24 (changes requested):
Almost good! Try to pass all tests, if you have any issues with it - feel free to text your fe_chat

1. `alphabetically`-button should be blue
image

src/App.tsx
```diff
@@ -16,34 +17,92 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  enum Sort {
+    None = '',
+    Alphabetic = 'alphabetic',
+    Length = 'length',
+  }
+
+  const [sortMethod, setSortmethod] = useState<Sort>(Sort.None);
```
```suggestion
  const [sortMethod, setSortMethod] = useState<Sort>(Sort.None);
```

src/App.tsx
```diff
@@ -16,34 +17,92 @@ export const goodsFromServer = [
...
+  const [isReversed, changeReverse] = useState(false);
```
etojeDenys, line 27:
```suggestion
  const [isReversed, setIsReversed] = useState(false);
```
always try to name state as [state, setState]

etojeDenys (approved):

PR #1816, opened 2024-04-24
add_solution

metrakonon:
[DEMO](https://metrakonon.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,12 @@
 import React from 'react';
-import 'bulma/css/bulma.css';
-import './App.scss';
+import { useState } from 'react';
 
-export const goodsFromServer = [
+import 'bulma/css/bulma.css';
+
+import './App.scss';
+import { GoodsList } from './Components/GoodsList/GoodsList';
+
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,37 +19,92 @@ export const goodsFromServer = [
...
+  Alphabetically = 'Alphabetically',
+  Length = 'Length',
+}
+
+enum ButtonType {
+  Alphabetically = 'Sort alphabetically',
+  Length = 'Sort by length',
+  Reverse = 'Reverse',
+  Reset = 'Reset',
+}
```
alexander-ignatow, line 32: You are using it only for the button text, wich looks like an overkill

src/App.tsx
```diff
@@ -15,37 +19,92 @@ export const goodsFromServer = [
...
+
+  if (reverse) {
+    sortedGoods = sortedGoods.reverse();
+  }
+
+  return sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sort, setSort] = useState<SortType | ''>('');
```
alexander-ignatow, line 62: Why not to allow '' be a valid sort type?
Alternatively, you may add `None` for an unsorted mode

src/App.tsx
```diff
@@ -15,37 +19,92 @@ export const goodsFromServer = [
...
+  const [reverse, setReverse] = useState<boolean>(false);
```
alexander-ignatow, line 63: It is ok just to 
```suggestion
  const [reverse, setReverse] = useState(false);
```

TypeScript can guess the type

src/App.tsx
```diff
@@ -15,37 +19,92 @@ export const goodsFromServer = [
...
+          onClick={() => setSort(SortType.Length)}
+        >
+          {ButtonType.Length}
         </button>
 
-        <button type="button" className="button is-warning is-light">
-          Reverse
+        <button
+          type="button"
+          className={`button is-warning ${!reverse && 'is-light'} `}
```
alexander-ignatow, line 87: For the future - it is the perfect case for `classnames`

alexander-ignatow (changes requested):
Looks good - just minor polishes are needed

src/App.tsx
```diff
@@ -1,8 +1,13 @@
 import React from 'react';
+import { useState } from 'react';
+
 import 'bulma/css/bulma.css';
+import cn from 'classnames';
+
 import './App.scss';
+import { GoodsList } from './Components/GoodsList/GoodsList';
```
mrkovaliv, line 8: improve react imports order

https://levelup.gitconnected.com/how-to-sort-imports-in-react-project-550f5ce70cbf

src/App.tsx
```diff
@@ -15,37 +20,88 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sort !== '' || reverse) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
```
mrkovaliv, line 94: move on this logic to handler func

mrkovaliv (approved):
Nice work 👍 👍 👍

PR #1817, opened 2024-04-25
adding task solution

eduvinagre:
[DEMO LINK](https://eduvinagre.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,37 +15,90 @@ export const goodsFromServer = [
...
+  const [sortedState, setSortedState] = useState<number>(0);
+
+  const reset = () => {
+    setList(goodsFromServer);
+    setReverseList(false);
+    setSortedState(0);
+  };
+
+  const sortedGoods = (goods: string[]) => {
+    setSortedState(1);
```
pedro-ruas, line 30: Eduardo, I believe you missed one point of the [Checklist](https://github.com/mate-academy/react_list-of-goods-hooks#:~:text=Use%20enum%20SortType%20instead%20of%20string%20literals%20for%20possible%20sorting%20options%20(including%20the%20default%20one)) 

You should use an Enum to test the sorting types instead of numbers

src/App.tsx
```diff
@@ -15,37 +15,90 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortedState !== 0 || reverseList) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => reset()}
```
pedro-ruas, line 86: This arrow function ins't necessary, you can just pass "reset" without calling it:
` onClick={reset} `

src/App.tsx
```diff
@@ -15,37 +15,90 @@ export const goodsFromServer = [
...
+          onClick={() => setList(sortedGoods(list))}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={`button is-success ${sortedState === 2 ? '' : 'is-light'}`}
+          onClick={() => setList(sortedGoodsLength(list))}
```
pedro-ruas, line 69: Instead of using setList with arrow functions for each button, you could create a Handle Function that makes the test for you and decides which sorting method to use.
Then, pass the handle function here: `onClick={handleSorting}`

src/App.tsx
```diff
@@ -15,37 +15,90 @@ export const goodsFromServer = [
...
+    const result = [...goods].sort((a, b) => a.localeCompare(b));
+
+    if (reverseList) {
+      return result.reverse();
+    }
+
+    return result;
+  };
+
+  const sortedGoodsLength = (goods: string[]) => {
```
pedro-ruas, line 40: The state variables are available for every function inside a component.
In this case, you don't need to create every sorting function receiving "goods" as parameter, because everytime you call it, you are passing the value of "goods" as the state value for "list".
You can simply remove the parameters from your sorting functions and let them access the state direcly

pedro-ruas (changes requested):
Great code! But you missed a checklist item and some small things can be improved

eduvinagre:
Fixed it!

jv-aquino (approved):
Great job Edu! The only suggestion i'd give is that the buttons could be all a single component with 1 or 2 props to change them, but this is more of a React encapsulation thing that we will always be practicing, for now, awesome TS and Hooks code!

eduvinagre:
@jv-aquino Nice observation. I'll take them to my future implementations! Tks, mate!

PR #1821, opened 2024-04-28
add task solution

JohnyJDX:
[DEMO LINK](https://JohnyJDX.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
+import cn from 'classnames';
+import React, { useState } from 'react';
 import './App.scss';
 
 export const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,35 +16,98 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  ByAlphabet = 'alphabet',
+  ByLength = 'length',
+}
+
+type SortBy = {
+  sortBy: SortType | string;
```
kinqbert, line 25: I believe you can write `sortBy: SortType | '';` instead to make code safer

src/App.tsx
```diff
@@ -15,35 +16,98 @@ export const goodsFromServer = [
...
+
+  if (isReverse) {
+    prepareGoods.reverse();
+  }
+
+  return prepareGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortBy, setSortBy] = useState('');
```
kinqbert, line 53: I believe you should add generic type for this state, because right now you can assign basically any string into it.

kinqbert (approved):
Great job! Check comments below for some inprovements! **Love your solution!**👍

src/App.tsx
```diff
@@ -15,35 +16,99 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  Alphabet = 'alphabet',
+  Length = 'length',
+  Default = '',
+}
+
+type SortBy = {
+  sortBy: SortType | string;
```
alexander-ignatow, line 26: I would add, why do you need `| string` or even `| ''`?

You already have all valid sort cases covered in `SortType` enum

src/App.tsx
```diff
@@ -15,35 +16,99 @@ export const goodsFromServer = [
...
+            type="button"
+            className="button is-danger is-light"
+          >
+            Reset
+          </button>
+        )}
       </div>
 
       <ul>
         <ul>
```
alexander-ignatow, line 106: Why you are nesting `<ul>` within another `<ul>` ?

alexander-ignatow (changes requested):
Good, just some minor changes are needed

src/App.tsx
```diff
@@ -15,36 +16,98 @@ export const goodsFromServer = [
...
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortBy || isReverse) && (
+          <button
+            onClick={() => {
```
mrkovaliv, line 93: move on this logic to handler func

mrkovaliv (approved):
👍 👍 👍

PR #1826, opened 2024-04-28
add task solution

lmuias:
[DEMO LINK](https://lmuias.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +17,92 @@ export const goodsFromServer = [
...
+    if (sortedBtn) {
+      newGoods.sort((good1, good2) => {
+        switch (sortedBtn) {
+          case SortBy.SORT_BY_ALPHABET:
+            return good1.localeCompare(good2);
+          case SortBy.SORT_BY_LENGTH:
+            return good1.length - good2.length;
+          default:
+            return 0;
+        }
```
volodymyr-soltys97, line 40:
```suggestion
        switch (sortedBtn) {
          case SortBy.SORT_BY_ALPHABET:
            return good1.localeCompare(good2);
            
          case SortBy.SORT_BY_LENGTH:
            return good1.length - good2.length;
            
          default:
            return 0;
        }
```

src/App.tsx
```diff
@@ -15,36 +17,92 @@ export const goodsFromServer = [
...
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortedBtn || isReverse) && (
+          <button
+            onClick={() => {
+              setSortedBtn('');
+              setIsReverse(false);
+            }}
```
volodymyr-soltys97, line 91: Move this logic from jsx to the helper reset function and use it here
```suggestion
            onClick={reset}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #1828, opened 2024-04-29
solution

GrigoryyKach:
[DEMO LINK](https://GrigoryyKach.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


PrytulaAlexandr (approved):
Good job)👍

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+  if (reversed) {
+    res.reverse();
+  }
+
+  return res;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
+  const [reversed, setReversed] = useState<boolean>(false);
```
alexander-ignatow, line 54: It is good to use typing like this, but in most cases TS can guess the type from the default value:

```suggestion
  const [sortField, setSortField] = useState(SortType.Default);
  const [reversed, setReversed] = useState(false);
```

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+          onClick={() => setSortField(SortType.Length)}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn('button is-warning', { 'is-light': !reversed })}
+          onClick={() => setReversed(!reversed)}
```
alexander-ignatow, line 92: Slightly better way is

```suggestion
          onClick={() => setReversed(value => !value)}
```

But no need to fix this

alexander-ignatow (approved):
Very good!

Please see my comments for the minor future updates

PR #1829, opened 2024-04-29
add solution

yanamomot:


src/App.tsx
```diff
@@ -15,35 +16,97 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortField {
+  Name = 'name',
+  Length = 'length',
+  NoField = '',
```
AYupko, line 22: I guess "Default" property would be clearer than NoField.

src/App.tsx
```diff
@@ -15,35 +16,97 @@ export const goodsFromServer = [
...
+}
+
+type SortProps = {
+  sortBy: SortField;
+  reversed: boolean;
+};
+
+function getPreparedGoods(products: string[], { sortBy, reversed }: SortProps) {
```
AYupko, line 30: Make sense to create different TS file and put the function in it.

src/App.tsx
```diff
@@ -15,35 +16,97 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortField {
+  Name = 'name',
```
AYupko, line 20: We are sorting not by name, but by alphabet. Not very proper naming.

AYupko (approved):
In general, good job. 
Just some notes about naming.

P.S. For convenience, put DEMO link in pull request description

src/App.tsx
DanilWeda: Agreed. ( not critical)

src/App.tsx
DanilWeda: Agreed! Please, move func to a separate file.

src/App.tsx
DanilWeda: Agreed! Need to fix.

DanilWeda (changes requested):
Good job. 
Please check some small improvements. 
Thanks)

src/App.tsx
```diff
@@ -16,34 +19,62 @@ export const goodsFromServer = [
...
+    reversed,
+  });
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${classNames({ 'is-light': sortBy !== SortField.Name })}`}
```
alexander-ignatow, line 35: just use following:
```suggestion
          className={classNames('button', 'is-info', { 'is-light': sortBy !== SortField.Name })}
```

alexander-ignatow (approved):
Good job 👍 

Please not following for the future: it is more common to name files with types `types.ts` (plural) - as most often you going to have multiple types
Also, see the comment in the code

PR #1835, opened 2024-04-29
add task solution

Voloshyn-Oleg:
https://Voloshyn-Oleg.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +17,102 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  ALPHABETICALLY = 'alphabetically',
+  BY_LENGTH = 'by_length',
+  REVERSE = 'reverse',
```
sevenlays, line 23: Good Job! My little cooment is: You have state isReverse, you can toggle that. Why make REVERSE property? I dont get it)

sevenlays (approved):

src/App.tsx
```diff
@@ -15,36 +17,102 @@ export const goodsFromServer = [
...
+  DEFAULT = 'default',
+}
+
+const getPreparedGoods = (
+  goods: string[],
+  currentField: SortType,
+  isReverse: boolean,
```
DanilWeda, line 30: You don`t need isReverce. You can use currentField and check SortType.REVERSE as in other cases.

DanilWeda (changes requested):
Good job. 
Check a small comment. 
Thanks)

src/App.tsx
```diff
@@ -15,36 +17,101 @@ export const goodsFromServer = [
...
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortType.DEFAULT);
+  const [isReverse, setIsReverse] = useState(false);
+
+  function handleSortClick(newSortField: SortType) {
+    if (newSortField === sortField) {
+      setIsReverse(isReverse);
+    } else {
+      setSortField(newSortField);
+    }
```
alexander-ignatow, line 57: Not sure you need to handle it like this - but lets keep it

alexander-ignatow (approved):
👍

PR #1848, opened 2024-04-30
add task solution

ArtemDemeniev:
https://artemdemeniev.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+const goodsFromServer: string[] = [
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +16,104 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const CONST_SORT_VALUE = 'alph';
+const CONST_LENGTH_VALUE = 'length';
+
+interface GoodProps {
+  sortField: string;
+  reverse: boolean;
+}
```
kinqbert, line 25: [Non critical] IMHO it is better to call it SortProps or something ike that. as GoodProps sound like these are some properties for each good (like quantity, category or etc.)

kinqbert (approved):
Great job! Everything seems to work as expected and code looks clean and easily readable. **Love your solution! 👍**

PR #1850, opened 2024-05-01
sonic1

yurii-filiushkin:
[DEMO LINK](https://yurii-filiushkin.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { visibleGoods } from './helpers/visibleGoods';
+import { SortType } from './types/SortTypes';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -16,34 +19,76 @@ export const goodsFromServer = [
...
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          onClick={() => {
+            if (sortBy !== SortType.length) {
+              return setSortBy(SortType.length);
+            }
+          }}
```
alexander-ignatow, line 50: In your case the check is redundant

```suggestion
          onClick={() => setSortBy(SortType.length)}
```

If sortBy does not change, simply nothing happens

src/App.tsx
```diff
@@ -16,34 +19,76 @@ export const goodsFromServer = [
...
+          })}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          onClick={() =>
+            reverse === false ? setReverse(true) : setReverse(false)
+          }
```
alexander-ignatow, line 62: Good, but a better form is:

```suggestion
          onClick={() => setReverse(value => !value)}
```

This is not critical though

src/App.tsx
```diff
@@ -16,34 +19,76 @@ export const goodsFromServer = [
...
+            type="button"
+            className="button is-danger is-light"
+          >
+            Reset
+          </button>
+        )}
       </div>
 
       <ul>
         <ul>
```
alexander-ignatow, line 86: ul inside ul!

src/App.tsx
```diff
@@ -16,34 +19,76 @@ export const goodsFromServer = [
...
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {resetButton && (
+          <button
+            onClick={() => {
+              setSortBy(SortType.none);
+              setReverse(false);
+            }}
```
alexander-ignatow, line 76: I suggest you create separate function (ex. `reset` and refactoring this place)

```suggestion
            onClick={reset}
```

alexander-ignatow (changes requested):
Good, but please you have several issues with this code.
Some are not critical, some a bit worse

Separately - please consider correct file naming - use `tsx` extension only, when you have JSX notation inside the file, otherwise please use just `ts`

You need to rename two files

src/App.tsx
yurii-filiushkin: Thanks for your comments.
this solve is cool (onClick={() => setReverse(value => !value)})

alexander-ignatow (approved):
👍

PR #1851, opened 2024-05-01
solution

Olek-91:
[DEMO LINK](https://Olek-91.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { SortTypes } from './Types/SortTypes';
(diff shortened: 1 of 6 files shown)


iAlfreX (approved):
Nice work!

src/App.tsx
```diff
@@ -15,35 +19,69 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const goods: Good[] = goodsFromServer.map((name, id) => ({ name, id }));
```
alexander-ignatow, line 22: It is OK to skip types sometimes - TS is able to calculate them
```suggestion
const goods = goodsFromServer.map((name, id) => ({ name, id }));
```

alexander-ignatow (approved):
You have several types in separate files, but it is more common to have a single file `types.ts` - at least, when you have a few small types, like in your case.

PR #1853, opened 2024-05-01
add soluion

stepanyukvova200:
[DEMO LINK](https://stepanyukvova200.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,36 +17,115 @@ export const goodsFromServer = [
...
+      return good1.length - good2.length;
+    });
+  }
+
+  if (fieldsToSort.includes(SortFields.reverse)) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
```
ilyasolo881, line 50: Nice work! Good solution

Have alternative shorter solution:

const getPreparedGoods = (
  goods: string[],
  sortType: SortType | null,
  reverse: boolean,
): string[] => {
  const preparedGoods = [...goods];

  if (sortType === SortType.ABC) {
    preparedGoods.sort((good1, good2) => good1.localeCompare(good2));
  } else if (sortType === SortType.LENGTH) {
    preparedGoods.sort((good1, good2) => good1.length - good2.length);
  }

  if (reverse) {
    preparedGoods.reverse();
  }

  return preparedGoods;
};

src/App.tsx
```diff
@@ -15,36 +17,115 @@ export const goodsFromServer = [
...
+        field === SortFields.lengthy
+          ? [...sortField, field].filter(
+              currentField => currentField !== SortFields.alphabetically,
+            )
+          : [...sortField, field].filter(
+              currentField => currentField !== SortFields.lengthy,
+            ),
+      );
+    }
+  }
```
ilyasolo881, line 75: It works, but look very scary, may be re-write on if else :)

PR #1858, opened 2024-05-02
react_list-of-goods-hooks

ksstassy:
[DEMO LINK](https://ksstassy.github.io/react_list-of-goods-hooks/)


lerastarynets (approved):
You're on the right track!

PR #1862, opened 2024-05-05
add task solution

MakhmudKerimov:
[DEMO LINK](https://MakhmudKerimov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,6 +8,7 @@ import './App.scss';
 enum SortType {
   Alphabetical = 'alphabetical',
   Length = 'length',
+  Default = '',
 }
 
 export const goodsFromServer: string[] = [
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +23,91 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const getPreparedGoods = (
+  goods: string[],
+  sortType: SortType | null,
```
etojeDenys, line 28:
```suggestion
  sortType: SortType,
```

src/App.tsx
```diff
@@ -15,36 +23,91 @@ export const goodsFromServer = [
...
+  reverse: boolean,
+): string[] => {
+  const preparedGoods = [...goods];
+
+  if (sortType === SortType.Alphabetical) {
+    preparedGoods.sort((good1, good2) => good1.localeCompare(good2));
+  } else if (sortType === SortType.Length) {
+    preparedGoods.sort((good1, good2) => good1.length - good2.length);
+  }
```
etojeDenys, line 37: you check different values of sortType so better to use a switch case here instead of if/else

src/App.tsx
```diff
@@ -15,36 +23,91 @@ export const goodsFromServer = [
...
+
+  if (reverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType | null>(null);
```
etojeDenys, line 47:
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.Default);
```

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1869, opened 2024-05-10
Develop

mfcastilho:
[DEMO LINK](https://mfcastilho.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import React, { useState } from 'react';
+import { GoodList } from './GoodList/index';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -16,36 +17,82 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [sortedState, setSortedState] = useState<number>(0);
+  const [reverseList, setReverseList] = useState<boolean>(false);
+  const [changedGoods, setChangedGoods] = useState<string[]>(goodsFromServer);
+
+  const handleAlphaSort = (): void => {
+    if (sortedState !== 1) {
+      setSortedState(1);
```
jv-aquino, line 26: it's not a good idea to use numbers to reference multiple types of sorting or things that are not related to numbers such as Strings

that's because if someone were to read your code they wouldn't understand which state relates to which sort method without further reading the entire code

jv-aquino (changes requested):
Good job Mario, there's just a minor thing that could be improved in terms of code readability:

EdPirro (approved):
Great job, @mfcastilho! Keep it up! 🚀

PR #1871, opened 2024-05-10
Solution

ropanefarias:
[DEMO LINK](https://ropanefarias.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +17,106 @@ export const goodsFromServer = [
...
+  Length = 'length',
+  Default = 'default',
+}
+
+type SortingFunction = (a: string, b: string) => number;
+
+const SORTING_FUNCTIONS: { [key in SortType]: SortingFunction } = {
+  [SortType.Alpha]: (a, b) => a.localeCompare(b),
+  [SortType.Length]: (a, b) => a.length - b.length,
+  [SortType.Default]: () => 0,
```
jv-aquino, line 31: you could also have no function in the default case

src/App.tsx
```diff
@@ -15,36 +17,106 @@ export const goodsFromServer = [
...
+            'is-light': sortField !== SortType.Alpha,
+          })}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          onClick={() => sortBy(SortType.Length)}
```
jv-aquino, line 87: a better practice would be to add a handler for each SortType - in general you should always have plain functions passed to JS events to improve mantainability

jv-aquino (changes requested):
Great job Rodrigo, your Typescript is for real very good, you did it in a very unique and good way! There's just some comments I left and a Github warning or error that you should fix so I can approve it!

src/App.tsx
```diff
@@ -15,36 +17,105 @@ export const goodsFromServer = [
...
+  [key in Exclude<SortType, SortType.Default>]: SortingFunction;
+} = {
+  [SortType.Alpha]: (a, b) => a.localeCompare(b),
+  [SortType.Length]: (a, b) => a.length - b.length,
+};
+
 export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<string[]>(goodsFromServer);
+  const [sortField, setSortField] = useState<SortType>(SortType.Default);
+  const [isReversed, setIsReversed] = useState<boolean>(false);
```
EdPirro, line 38: No need to explicitly set types that can be implied from values; 
Typescript internally handles types; simply use `useState(valueWithImplicitType)`

EdPirro (approved):
Great job, @ropanefarias! Keep it up! 🚀

PR #1884, opened 2024-05-22
task solution

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,12 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+
+interface SortParams {
+  sortField: string;
+  reverse: boolean;
+}
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,36 +21,98 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const SORT_FIELD_ALPHABETICALLY = 'alphabetically';
```
yevhenii-pyl, line 24: Valid approach, but cleaner way to do it  will be to create a type for it, like:

```
enum SortType {
  ALPHABETICALLY = 'alphabetically',
  BY_LENGTH = 'length',
}
```

Or a simple object, in case you work with js without ts.

src/App.tsx
```diff
@@ -15,36 +21,98 @@ export const goodsFromServer = [
...
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reverse, setReverse] = useState(false);
+  const visibleGoods = getPreparedGood(goodsFromServer, { sortField, reverse });
+  const resetSort = () => {
```
yevhenii-pyl, line 59: 👍

src/App.tsx
```diff
@@ -15,36 +21,98 @@ export const goodsFromServer = [
...
+    return preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reverse, setReverse] = useState(false);
+  const visibleGoods = getPreparedGood(goodsFromServer, { sortField, reverse });
```
yevhenii-pyl, line 58: 💪

src/App.tsx
```diff
@@ -15,36 +21,98 @@ export const goodsFromServer = [
...
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortField || reverse) && (
+          <button
+            type="button"
+            className={cn('button is-danger', {
+              'is-light': sortField || reverse,
+            })}
+            onClick={() => resetSort()}
```
yevhenii-pyl, line 103: `onClick={resetSort}`

You can use a shortcut here, but it's just a matter of preference.

yevhenii-pyl (approved):
Solid!  🥇 

The only thing to pay attention is using enums || objects to gather such expandable (possibly, in future) and reusable data sets like sorting types.

PR #1894, opened 2024-05-27
solution

getchards:


src/App.tsx
```diff
@@ -15,35 +15,101 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  name = 'name',
+  length = 'length',
+}
+
 export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<string[]>(goodsFromServer);
+  const [sortField, setSortField] = useState<SortType | ''>('');
```
```suggestion
  const [sortField, setSortField] = useState<SortType>(SortType.default);
```

src/App.tsx
```diff
@@ -15,35 +15,101 @@ export const goodsFromServer = [
...
+          return good1.length - good2.length;
+        default:
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      sortedGoods.reverse();
+    }
+
```
etojeDenys: move this logic to some function (getVisibleGoods)

and just call it
`const visibleGoods = getVisibleGoods(...........)`

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+            return good1.length - good2.length;
+          default:
+            return 0;
+        }
+      });
+
+      if (isReversed) {
+        sortedGoods.reverse();
+      }
+
```
etojeDenys, line 70: move all this logic to some function (getVisibleGoods)

and then just call it
`const visibleGoods = getVisibleGoods(.......)`
and use this var instead of the  setVisibleGoods state

etojeDenys (changes requested):
feel free ti ask for some help in the chat

src/App.tsx
```diff
@@ -15,36 +16,120 @@ export const goodsFromServer = [
...
+  useEffect(() => {
+    const sortedVisibleGoods = getVisibleGoods(
+      visibleGoods,
+      sortField,
+      isReversed,
+    );
+
+    setVisibleGoods(sortedVisibleGoods);
+  }, [visibleGoods, sortField, isReversed]);
```
TarasHoliuk, line 65: you don't need separate state for visibleGoods

General rule: if you can "calculate" your value based on other state or props values you don't need a separate state for it. Because it will be recalculated every time when state/props change - during re-render

So:

```suggestion
    const visibleGoods = getVisibleGoods(
      goodsFromServer,
      sortField,
      isReversed,
    );
```

src/App.tsx
```diff
@@ -15,36 +16,120 @@ export const goodsFromServer = [
...
+  };
+
+  const reset = () => {
+    setVisibleGoods(goodsFromServer);
+    setIsReversed(false);
+    setSortField(SortType.default);
+  };
+
+  const reverseList = () => {
+    setIsReversed(!isReversed);
```
TarasHoliuk, line 82:
```suggestion
    setIsReversed((currentIsReversed) =>  !currentIsReversed);
```
It's better to use functional setState in such cases: [updating-state-based-on-the-previous-state](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state)

src/App.tsx
```diff
@@ -15,36 +16,120 @@ export const goodsFromServer = [
...
+          className={classNames('button', 'is-success', {
+            'is-light': !isReversed,
+          })}
+          onClick={reverseList}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
+        <button
```
TarasHoliuk, line 118: check tests and task requirements: this button shouldn't be visible if there is nothing to reset (it's default order now)

src/App.tsx
```diff
@@ -15,36 +16,120 @@ export const goodsFromServer = [
...
+          onClick={sortAlphabetically}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={classNames('button', 'is-success', {
+            'is-light': sortField !== SortType.length,
```
TarasHoliuk, line 101: check buttons colors in [the working version](https://mate-academy.github.io/react_list-of-goods/)

TarasHoliuk (changes requested):

PR #1915, opened 2024-06-17
solution

myemets:


src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
...
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(isReversed || sortType) && (
+          <button
+            onClick={() => reset()}
```
VitaliyBondarenko1982, line 103:
```suggestion
            onClick={reset}
```
it is the same

src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
...
+          className={classNames('button is-info', {
+            'is-light': sortType !== SortType.Alphabet,
+          })}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          onClick={() => setSortType(SortType.Length)}
```
VitaliyBondarenko1982, line 82: suggest to make handler factory in component and use for both sort buttons.

```
const makeHandleSort = (sortType: SortType) => () => setSortType(sortType)

```
```suggestion
          onClick={makeHandleSort(SortType.Length)}
```

src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
...
+          className={classNames('button is-success', {
+            'is-light': sortType !== SortType.Length,
+          })}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          onClick={() => setIsReversed(!isReversed)}
```
VitaliyBondarenko1982, line 92: it is bad practice to make anonymous arrow functions in jsx.
make handler in component and pass reference on it here (as for reset button).

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check comments and provide demo link in PR description.

anastasiiavorobiova (approved):
Awesome work!

PR #1919, opened 2024-06-20
add task solution

romanyshyns:
[DEMO LINK](https://romanyshyns.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +16,99 @@ export const goodsFromServer = [
...
-        </button>
+        {(sortField !== '' || reverse) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField('');
+              setReverse(false);
+            }}
+          >
```
anastasiiavorobiova, line 100:
```suggestion
            onClick={reset}
          >
```
Could you consider moving this function from the JSX? It would allow us to reuse this function and optimize it with the useCallback hook in the future

anastasiiavorobiova (approved):
Great work!

PR #1925, opened 2024-06-22
enum type

MariCodec:
[DEMO LINK](https://MariCodec.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,37 +14,85 @@ export const goodsFromServer = [
...
   'Garlic',
 ];
+enum SortType {
+  None = '',
+  Alphabetical = 'alphabetical',
+  Length = 'length',
+}
 
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType>(SortType.None);
```
LitviVit, line 24: In this case the explicit generic parameter is not strictly necessary because TS can infer the type from the initial value. BUT it is good practice to include it for clarity and type safety👍🏻

src/App.tsx
```diff
@@ -14,37 +14,85 @@ export const goodsFromServer = [
...
+  }
+
+  const reversedGoods = () => {
+    setIsReversed(!isReversed);
+  };
+
+  const resetState = () => {
+    setSortType(SortType.None);
+    setIsReversed(false);
+  };
```
LitviVit, line 46: So, if you decide to add explicit generic type, maybe you should add types here too? It's not a mistake and there is no need any changes, because TS can infer the correct types for you, especially in simple cases like this. Just note😄
```suggestion
  const reversedGoods: () => void = () => {
    setIsReversed(!isReversed);
  };

  const resetState: () => void = () => {
    setSortType(SortType.None);
    setIsReversed(false);
  };
```

LitviVit (approved):
Awesome! Well done

src/App.tsx
```diff
@@ -14,37 +14,85 @@ export const goodsFromServer = [
   'Jam',
   'Garlic',
 ];
+enum SortType {
+  None = '',
```
VenomMB, line 18: Better 
```suggestion
  Default = 'default',
```

src/App.tsx
```diff
@@ -14,37 +14,85 @@ export const goodsFromServer = [
...
+  const [sortType, setSortType] = useState<SortType>(SortType.None);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const visibleGoods = [...goodsFromServer];
+
+  if (sortType === SortType.Alphabetical) {
+    visibleGoods.sort();
+  } else if (sortType === SortType.Length) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
```
VenomMB, line 33: Please, prefer switch/case over multiple if/else

It may not be critical on two possible SortTypes, but still readability of the code suffers.

src/App.tsx
```diff
@@ -14,37 +14,85 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map((good, index) => {
+          const uniqueKey = `${index}`;
+
+          return (
+            <li key={uniqueKey} data-cy="Good">
+              {good}
+            </li>
```
VenomMB, line 93: good practice use uuid library

VenomMB (approved):
Good Job!!!
Leave some comments

PR #1927, opened 2024-06-22
add solution task react_list-of-goods-hooks

seriuksergii:
[DEMO LINK](https://seriuksergii.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
-import 'bulma/css/bulma.css';
+import { useState } from 'react';
 import './App.scss';
+import cn from 'classnames';
+import 'bulma/css/bulma.css';
+import React from 'react';
 
 export const goodsFromServer = [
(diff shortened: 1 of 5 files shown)


yaroslav1177 (approved):
Good job!

src/App.tsx
```diff
@@ -15,36 +17,106 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn('button is-warning', {
+            'is-light': !reverse,
+          })}
+          onClick={() => setReverse(!reverse)}
```
alexander-ignatow, line 98: No need to fix in your case, but it is always better to use the following:

```suggestion
          onClick={() => setReverse(value => !value)}
```

alexander-ignatow (approved):
Good job 👍 !

PR #1932, opened 2024-06-23
add solution to the react_list-of-goods-hooks task on typescript

lesdyman:
Hi there! Here is my hard-won solution for the react_list-of-goods-hooks task on typescript. Hope I did everything as it was supposed to be. You shall find the demo link below. Cheers! 

[DEMO LINK](https://lesdyman.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,35 +17,103 @@ export const goodsFromServer = [
...
+
+    if (isReversed && sortingType !== SortType.RESET) {
+      sortedList.reverse();
+    }
+
+    setVisibleList(sortedList);
+    setActualSortType(sortingType);
+  };
+
+  const reverseList = () => {
```
FaiHamid, line 60: this part repeated, you can change your function for only one check isReversed

FaiHamid (approved):
Well done!!! 
I add only a couple of recommendations)

src/App.tsx
```diff
@@ -16,35 +17,103 @@ export const goodsFromServer = [
...
+    const reversedList = [...visibleList].reverse();
+
+    setVisibleList(reversedList);
+    setIsReversed(prev => !prev);
```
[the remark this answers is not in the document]
lesdyman: Good point! I'll change this piece a little)

src/App.tsx
lesdyman: I'm not sure I fully understand this one. Could you please explain a little bit more, would you kindly?  I think you are seeing something important at this point. And I'm not there yet))

alexander-ignatow: BTW - it is ok to have separate functions just to set the state, but performing actual manipulations on data is another thing

lesdyman:
@alexander-ignatow Understood! Thanks for review! I've changed my solution a little bit). Can't tell this was walk in a park for me. I had to use some advise at the Q&A session. But now code looks better I think)

roman-mirzoian (approved):
Well done!

PR #1933, opened 2024-06-23
add task solution

V-Shut:
[DEMO LINK](https://v-shut.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import React, { useState } from 'react';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
+      ...trackStatus,
+      sortType: SortType.Length,
+      reversed: reversed,
+    });
+  };
+
+  const handleReverse = () => {
+    setTrackStatus({
+      ...trackStatus,
+      reversed: !reversed,
```
lesdyman, line 75: I wasn't sure about defaultState constant you use. But now I get it)). This definitely may be a way to solve the question. But also you could create another useState for a reversed state)

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getGoods().map(good => (
+          <li data-cy="Good" key={good}>
+            {good}
+          </li>
+        ))}
```
lesdyman, line 126: At the and of the day your code looks so pure and neat. Love it))

lesdyman (approved):
Nicely done! I've left some comments. But what I meant is that there are so many ways to solve different issues. I like how neat your code looks like though

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
+  const { reversed, sortType } = trackStatus;
+
+  const getGoods = () => {
+    const goods = [...goodsFromServer];
+
+    if (sortType === SortType.Alphabetical) {
+      goods.sort((a, b) => a.localeCompare(b));
+    } else if (sortType === SortType.Length) {
+      goods.sort((a, b) => a.length - b.length);
+    }
```
alexander-ignatow, line 47: Please, prefer `switch/case` over multiple `if/else`

It may not be critical on two possible SortTypes, but still readability of the code suffers.

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  Default = 'default',
+  Alphabetical = 'alphabetical',
+  Length = 'length',
+}
+
+interface TrackStatus {
```
alexander-ignatow, line 25: The name of the type is misleading.

Maybe `Filter`, `FilterProps`, `SortFields`, etc?

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
+
+    return goods;
+  };
+
+  const handleSortAlphabetically = () => {
+    setTrackStatus({
+      ...trackStatus,
+      sortType: SortType.Alphabetical,
+      reversed: reversed,
+    });
```
alexander-ignatow, line 61: why do you re-set `reserved` field, if you have already done it with `...trackStatus`?

```suggestion
    setTrackStatus({
      ...trackStatus,
      sortType: SortType.Alphabetical,
          });
```

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
+  };
+
+  const handleSortByLength = () => {
+    setTrackStatus({
+      ...trackStatus,
+      sortType: SortType.Length,
+      reversed: reversed,
+    });
+  };
```
alexander-ignatow, line 70: Same as above:

```suggestion
  const handleSortByLength = () => {
    setTrackStatus({
      ...trackStatus,
      sortType: SortType.Length,
    });
  };
```

src/App.tsx
```diff
@@ -15,36 +16,114 @@ export const goodsFromServer = [
...
+
+  const handleReverse = () => {
+    setTrackStatus({
+      ...trackStatus,
+      reversed: !reversed,
+    });
```
alexander-ignatow, line 76: It changes nothing in your case, but it is better to use the following:

```suggestion
    setTrackStatus(value => ({
      ...value,
      reversed: !value.reversed,
    }));
```

alexander-ignatow (changes requested):
Good job 👍 , but some changes are needed

src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
+enum SortType {
+  Default = 'default',
+  Alphabetical = 'alphabetical',
+  Length = 'length',
+}
+
+interface FilterProps {
+  reversed: boolean;
+  sortType: SortType;
+}
```
VenomMB, line 28: Good practice move types to another file, only for types.

src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {getGoods().map(good => (
+          <li data-cy="Good" key={good}>
```
VenomMB, line 125: good practice use uuid library

VenomMB (approved):
Good Job!
Leave some comments

PR #1958, opened 2024-07-05
Solution

kyrylomanko:
[DEMO LINK](https://kyrylomanko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,37 +16,104 @@ export const goodsFromServer = [
...
-export const App: React.FC = () => {
+type Good = string;
+
+enum SortType {
+  NONE = '',
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+}
+
+const GoodList: React.FC<{ goods: Good[] }> = ({ goods }) => (
```
volodymyr-soltys97, line 27: It's a bad practice to create two components in the one file.
You don't need a separate component for this logic, or if you want to create one, you need to export it to a separate file

src/App.tsx
```diff
@@ -15,37 +16,104 @@ export const goodsFromServer = [
...
+  if (sortField) {
+    prepearedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+        default:
+          return 0;
+      }
```
volodymyr-soltys97, line 52:
```suggestion
      switch (sortField) {
        case SortType.ALPHABET:
          return good1.localeCompare(good2);
          
        case SortType.LENGTH:
          return good1.length - good2.length;
          
        default:
          return 0;
      }
```

src/App.tsx
```diff
@@ -15,37 +16,104 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {(sortField || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortType.NONE);
+              setIsReversed(false);
+            }}
```
volodymyr-soltys97, line 110: Move this logic from jsx to the helper function and use it here
```suggestion
            onClick={reset}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Good job 👍 
Consider will be better if you rename this file to `GoodList.tsx`
image

PR #1961, opened 2024-07-10
Develop

akryto:
- Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript) if needed.
- Use `enum SortType` instead of string literals for possible sorting options (including the default one)

https://akryto.github.io/react_list-of-goods-hooks/


demkoveronika (approved):
Good job))

PR #1969, opened 2024-07-19
add solution

timaantonuk:
[DEMO LINK](https://timaantonuk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,36 +16,99 @@ export const goodsFromServer = [
...
+      [...goodsList].sort(
+        (good1: string, good2: string) => good1.length - good2.length,
+      ),
+    );
+  };
+
+  const sortByReverse = (): void => {
+    setReversed(!reversed);
+    setGoodsList([...goodsList].reverse());
+  };
```
SanyaBratashchuk, line 54:
```suggestion
  const visibleGoods = goodsList.sort((a, b) => {
     switch(sortType) {
        here logic of sorting
     }
  }).
```

src/App.tsx
```diff
@@ -15,36 +16,99 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goodsList.map(good => (
+          <li data-cy="Good" key={generateRandomNumericId()}>
```
SanyaBratashchuk, line 108:
```suggestion
          <li data-cy="Good" key={good}>
```

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -16,35 +17,90 @@ export const goodsFromServer = [
...
-        </button>
+        {sortType !== '' || reversed ? (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={resetSort}
+          >
+            Reset
+          </button>
+        ) : null}
```
etojeDenys, line 95: simplfify it using the && instead of ?:

src/App.tsx
```diff
@@ -16,35 +17,90 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [goodsList, setGoodsList] = useState<string[]>([...goodsFromServer]);
+  const [sortType, setSortType] = useState<string>('');
+  const [reversed, setReversed] = useState<boolean>(false);
```
etojeDenys, line 22: we don't need to specify types for primitives 
```suggestion
  const [sortType, setSortType] = useState('');
  const [reversed, setReversed] = useState(false);
```

src/App.tsx
```diff
@@ -16,35 +17,90 @@ export const goodsFromServer = [
...
+      case 'reverse':
+        sortedGoods.reverse();
+        break;
+      default:
+        sortedGoods = [...goodsFromServer];
+    }
+
+    setSortType(type);
+    setGoodsList(sortedGoods);
+  };
```
etojeDenys, line 43:
```suggestion
  const getPreparedGoods = (sortType: string, reversed: boolean): void => {
    let sortedGoods = [...goodsList];

    switch (type) {
      case 'by name':
        sortedGoods.sort();
        break;
      case 'by length':
        sortedGoods.sort((good1, good2) => good1.length - good2.length);
        break;
      default:
        sortedGoods = [...goodsFromServer];
    }
    
    if(reverse) {
      sortedGoods.reverse();
    }

   return sortedGoods
  };
```

and now you can just call this function and use the results everywhere
`const visibleGoods = getPreparedGoods(sortType, reversed)`

src/App.tsx
```diff
@@ -16,35 +17,90 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [goodsList, setGoodsList] = useState<string[]>([...goodsFromServer]);
```
etojeDenys, line 20: after that you can remove this state

etojeDenys (changes requested):

PR #1976, opened 2024-08-01
Solution

vitorhvieira:
[DEMO LINK](https://vitorhvieira.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,37 +17,103 @@ export const goodsFromServer = [
...
 ];
 
+enum SortField {
+  DEFAULT = '',
+  ALPHABETIC = 'alphabetically',
+  LENGTH = 'length',
+}
+
+interface SortParams {
+  sortField: string;
```
pedro-ruas, line 27: You have declared an Enum for this, instead of passing the type as `string` you can use your declared Enum `sortField: SortField`, it is a better approach

src/App.tsx
```diff
@@ -15,37 +17,103 @@ export const goodsFromServer = [
...
+
+  if (reversedField) {
+    prepareGoods.reverse();
+  }
+
+  return prepareGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState(SortField.DEFAULT);
```
pedro-ruas, line 58: You can type `useState` too (not mandatory, but it is a good practice):
```
useState<SortField>(SortField.DEFAULT);
```

src/App.tsx
```diff
@@ -15,37 +17,103 @@ export const goodsFromServer = [
...
+  const [reversedField, setReversedField] = useState(false);
```
pedro-ruas, line 59: And here too 
```
useState<boolean>(false);
```

pedro-ruas (changes requested):
Really minimal improvements, your code looks great already!

pedro-ruas (approved):
Great!!!

PR #1980, opened 2024-08-04
add task solution

AndriiZakharenko:
[DEMO LINK](https://AndriiZakharenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ export const goodsFromServer = [
 enum SortType {
   length = 'length',
   alphabet = 'alphabet',
+  default = '',
 }
 
 function getPraparedGoods(goods: string[], sortField: string) {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,11 +22,11 @@ enum SortType {
   default = '',
 }
 
-function getPraparedGoods(goods: string[], sortField: string) {
-  const preparedGoods = [...goods];
+function sortGoods(goods: string[], sortField: SortType) {
+  const sortedGoods = [...goods];
 
   if (sortField) {
-    preparedGoods.sort((good1, good2) => {
+    sortedGoods.sort((good1, good2) => {
       switch (sortField) {
         case SortType.alphabet:
           return good1.localeCompare(good2);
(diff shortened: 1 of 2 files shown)


Teollan (approved):
Nicely done 🥇 !

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  length = 'length',
+  alphabet = 'alphabet',
+  default = '',
+}
+
+function getPraparedGoods(goods: string[], sortField: string) {
```
BudnikOleksii, line 25:
```suggestion
function sortGoods(goods: string[], sortField: SortType) {
```
It's not a string, you have enum, so use it. And you don't prepare goods, just sort them

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
...
+      }
+    });
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [reversed, setReversed] = useState(false);
```
BudnikOleksii, line 48: - The same, it's not a string
- Boolean variables should start from `is` or `has`

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
...
+  let visibleGoods = getPraparedGoods(goodsFromServer, sortField);
+
+  if (reversed) {
+    visibleGoods = visibleGoods.reverse();
+  }
```
BudnikOleksii, line 53:
```suggestion
  const sortedGoods = getPraparedGoods(goodsFromServer, sortField);
  const visibleGoods = isReversed ? sortedGoods.reverse() : sortedGoods;
```

src/App.tsx
```diff
@@ -15,36 +16,95 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setReversed(!reversed)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortField.length > 0 || reversed) && (
```
BudnikOleksii, line 88:
```suggestion
        {(sortField || reversed) && (
```
`sortField` is a string, so just check if it exists. And create a separate variable for this condition to avoid combining different logical operators in the same place

BudnikOleksii (changes requested):
Good work 👍 

BudnikOleksii (approved):
Nicely done 🔥

PR #1982, opened 2024-08-05
add solution

RimmaBuzuluk:
[DEMO LINK](https://RimmaBuzuluk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


MaxBorovique (approved):
Well done! As usual!

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  Alphabet = 'Sort alphabetically',
+  Length = 'Sort by length',
+}
+
+function goodInUse(
+  good: string[],
+  isReverse: boolean,
+  sortTypeFilter: string,
```
BudnikOleksii, line 27: It's not a string, use your enum `SortType`

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
...
+
+  if (isReverse) {
+    currentGoods.reverse();
+  }
+
+  return currentGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortTypeFilter, setSortTypeFilter] = useState('');
```
BudnikOleksii, line 52: The same

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
...
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {sortTypeFilter === '' && isReverse === false ? (
+          ''
+        ) : (
```
BudnikOleksii, line 95: You don't need to compare with an empty string, just check if a value exists or not. And the same about comparing with boolean variables
https://developer.mozilla.org/en-US/docs/Glossary/Falsy
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
...
 
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {goodsList.map(item => (
```
BudnikOleksii, line 111: What item is it?

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  Alphabet = 'Sort alphabetically',
+  Length = 'Sort by length',
+}
+
+function goodInUse(
```
BudnikOleksii, line 24: Function name should start from verb

src/App.tsx
```diff
@@ -15,35 +16,103 @@ export const goodsFromServer = [
...
+    currentGoods.reverse();
+  }
+
+  return currentGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortTypeFilter, setSortTypeFilter] = useState('');
+  const [isReverse, setIsReverse] = useState(false);
+  const goodsList: string[] = goodInUse(
```
BudnikOleksii, line 54: You don't need to name with `List` as we know it as it's an array. Could be `sortedGoods/preparedGoods` or something similar

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

BudnikOleksii (approved):
Good job 👍

PR #1996, opened 2024-08-07
done

NIKaragu:
[DEMO LINK](https://NIKaragu.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


Kateryna99 (approved):

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  SORT_BY_FIRST_LETTER = 'byLetter',
+  SORT_BY_GOOD_LENGTH = 'byLength',
+  RESET_SORTING = '',
```
BudnikOleksii, line 22: Sort is not only by the first letter. Also, properties should be in camelCase

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+}
+
+function getSortedGoods(
+  goodsList: string[],
+  sortField: SortType,
+  reversed: boolean,
+) {
+  let goodsListCopy = [...goodsList];
```
BudnikOleksii, line 30: Prefer const over let and you don't need to add "list" in name is it's an array and everyone understand it

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+        case SortType.SORT_BY_GOOD_LENGTH:
+          return good1.length - good2.length;
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (reversed) {
+    goodsListCopy = [...goodsListCopy].reverse();
```
BudnikOleksii, line 46:
```suggestion
    goodsListCopy.reverse();
```
You can just reverse it

BudnikOleksii (approved):
Good work 👍 
Approved and left a few suggestions for the future

PR #2002, opened 2024-08-09
add solution

YarikKedrych:


Yurii-Yaremko (approved):
You need add [DEMO LINK] 
Good job! Approve)

src/App.tsx
```diff
@@ -15,36 +17,95 @@ export const goodsFromServer = [
...
+
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState<boolean>(false);
+  const [sortType, setsortType] = useState<SortType | ''>('');
+
+  const sortedGoods: string[] = handleSort(sortType, isReversed);
+  const shouldShowResetButton =
+    sortType === SortType.SORT_ALPHABET ||
+    sortType === SortType.SORT_LENGTH ||
+    isReversed;
```
BudnikOleksii, line 55:
```suggestion
  const shouldShowResetButton = sortType  || isReversed;
```
Could be simplified

BudnikOleksii (changes requested):
Good work 👍 
Please, add a link to demo so a reviewer can check your solutioin

src/App.tsx
```diff
@@ -15,36 +17,92 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  SORT_ALPHABET = 'alphabet',
+  SORT_LENGTH = 'length',
+}
+
+const handleSort = (type: SortType | '', isReversed: boolean): string[] => {
```
DanilWeda, line 25: In this case better name getSortedGoods ( or something like that). 
Use the handle for events ( onClick, onChange, etc.)

DanilWeda, line 25: Also, you can move this function to a separate file, import it, and use it inside the component.

DanilWeda (approved):
Good job. 
I approved, but please check the comments about naming. 
Try to fix it. Many thanks)

PR #2015, opened 2024-08-14
solution

DanielLinker17:
https://DanielLinker17.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 import 'bulma/css/bulma.css';
 import './App.scss';
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import { ButtonsList } from './components/ButtonsList';
 import { GoodsList } from './components/GoodsList';
 
@@ -17,23 +17,41 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const getSortedGoods = (
+  goods: string[],
+  sortCriteria: string,
+  isReversed: boolean,
+): string[] => {
+  const sortedGoods = [...goods];
+
+  if (sortCriteria === 'alphabet') {
+    sortedGoods.sort((a, b) => a.localeCompare(b));
+  } else if (sortCriteria === 'length') {
+    sortedGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    sortedGoods.reverse();
+  }
+
+  return sortedGoods;
+};
+
 export const App: React.FC = () => {
-  const [goods, setGoods] = useState(goodsFromServer);
-  const [sortCriteria, setsortCriteria] = useState('');
+  const [sortCriteria, setSortCriteria] = useState('');
   const [isReversed, setIsReversed] = useState(false);
 
-  const handlesortCriteria = (action: string) => {
+  const handleSortCriteria = (action: string) => {
     switch (action) {
       case 'alphabet':
       case 'length':
-        setsortCriteria(action);
+        setSortCriteria(action);
         break;
       case 'reverse':
         setIsReversed(prev => !prev);
         break;
       case 'reset':
-        setsortCriteria('');
-        setGoods(goodsFromServer);
+        setSortCriteria('');
         setIsReversed(false);
         break;
       default:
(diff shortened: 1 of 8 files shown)


src/components/GoodsList.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+type GoodsListProps = {
+  goods: string[];
+};
+
+export const GoodsList: React.FC<GoodsListProps> = ({ goods }) => {
+  return (
+    <ul>
+      {goods.map(good => (
+        <Good goodName={good} key={createRandomId()} />
```
etojeDenys, line 12:
```suggestion
        <Good goodName={good} key={good} />
```
you can just use good as a key

src/components/Good.tsx
```diff
@@ -0,0 +1,7 @@
+type GoodProps = {
+  goodName: string;
+};
+
+export const Good: React.FC<GoodProps> = ({ goodName }) => {
+  return <li data-cy="Good">{goodName}</li>;
+};
```
etojeDenys, line 7: it's redundant to create so simple component, better to write it directly inside GoodsList

src/components/ButtonsList.tsx
```diff
@@ -0,0 +1,56 @@
...
+        name="reverse"
+        content="Reverse"
+        handleFilter={handleFilter}
+        filter={filter}
+        isReversed={isReversed}
+      />
+
+      {(filter || isReversed) && (
+        <Button
+          key={createRandomId()}
```
etojeDenys, line 46: we usually use key attr only when trying to map some elements, here it's redundant

src/components/Button.tsx
```diff
@@ -0,0 +1,36 @@
...
+  content,
+  handleFilter,
+  filter,
+  isReversed,
+}) => {
+  const buttonClass = classNames('button', {
+    'is-info': name === 'alphabet',
+    'is-success': name === 'length',
+    'is-warning': name === 'reverse',
+    'is-danger': name === 'reset',
```
etojeDenys, line 22: looks like better to create some helper function that will accept the name and return the className, you can use the switch/case there

src/components/Button.tsx
```diff
@@ -0,0 +1,36 @@
+import classNames from 'classnames';
+
+type ButtonProps = {
+  name: string;
+  content: string;
```
etojeDenys, line 5: you can just use the children prop 

https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children

src/components/Button.tsx
```diff
@@ -0,0 +1,36 @@
...
+    'is-info': name === 'alphabet',
+    'is-success': name === 'length',
+    'is-warning': name === 'reverse',
+    'is-danger': name === 'reset',
+    'is-light': name !== filter && !(name === 'reverse' && isReversed),
+  });
+
+  return (
+    <button
+      name={name}
```
etojeDenys, line 28: it's redundant attr for button

src/App.tsx
```diff
@@ -16,36 +18,55 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState(false);
+
+  const handleFilter = (action: string) => {
+    switch (action) {
+      case 'alphabet':
+        setFilter('alphabet');
+        break;
+      case 'length':
+        setFilter('length');
+        break;
```
etojeDenys, line 32:
```suggestion
    switch (action) {
      case 'alphabet':
      case 'length':
        setFilter(action);
        break;
```

src/App.tsx
```diff
@@ -16,36 +18,55 @@ export const goodsFromServer = [
...
+    } else if (filter === 'length') {
+      sortedGoods = sortedGoods.sort((a, b) => a.length - b.length);
+    }
+
+    if (isReversed) {
+      sortedGoods.reverse();
+    }
+
+    setGoods(sortedGoods);
+  }, [filter, isReversed]);
```
etojeDenys, line 60: try to rewrite it without using the state for goods

we can just 
```
const goods = getSortedGoods(goodsFromServer, filter, isReversed)
```
and move all logic to that function

src/App.tsx
```diff
@@ -16,36 +18,55 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState(goodsFromServer);
+  const [filter, setFilter] = useState('');
```
etojeDenys, line 22: it's not a filter, better to name it as `sortValue` or something like that

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Good job! 👍

PR #2017, opened 2024-08-17
add task solution

Pokhvalenkos:
[DEMO LINK](https://pokhvalenkos.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,7 +61,7 @@ export const App = () => {
   const visibleGoods = getPreparedGoods(goodsFromServer, sortField, isReversed);
   const showResetButton = sortField || isReversed;
 
-  function dangerHandler() {
+  function resetSort() {
     setSortField(SortType.None);
     setIsReversed(false);
   }
@@ -91,7 +91,7 @@ export const App = () => {
 
         <button
           type="button"
-          className={`button is-info ${isReversed ? '' : 'is-light'}`}
+          className={classNames(`button is-info`, { 'is-light': !isReversed })}
           onClick={() => setIsReversed(!isReversed)}
         >
           Reverse
@@ -101,9 +101,7 @@ export const App = () => {
           <button
             type="button"
             className="button is-danger is-light"
-            onClick={() => {
-              dangerHandler();
-            }}
+            onClick={resetSort}
           >
             Reset
           </button>


src/App.tsx
```diff
@@ -15,37 +17,98 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
+  sort_field_alphabet = 'alphabet',
+  sort_field_length = 'length',
+  sort_field_none = '',
```
TarasHoliuk, line 23:
```suggestion
  Alphabet = 'alphabet',
  Length = 'length',
  None = '',
```
use PascalCase for enum keys. Also, no need to duplicate SortField in each key - it's already clear from the enum name

src/App.tsx
```diff
@@ -15,37 +17,98 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setSortField(SortType.sort_field_length)}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-info ${isReversed ? '' : 'is-light'}`}
```
TarasHoliuk, line 91: use classNames here too

Also, in the working example this button isn't blue (not is-info class)

src/App.tsx
```diff
@@ -15,37 +17,98 @@ export const goodsFromServer = [
...
+          onClick={() => setIsReversed(!isReversed)}
```
TarasHoliuk, line 92: update state based on current value: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/App.tsx
```diff
@@ -15,37 +17,98 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {showResetButton && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
```
TarasHoliuk, line 101: it's better to move into a separate handler

TarasHoliuk (changes requested):
GJ 👍 
let's improve your code a bit

src/App.tsx
```diff
@@ -15,37 +17,100 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setSortField(SortType.Length)}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-info ${isReversed ? '' : 'is-light'}`}
```
volodymyr-soltys97, line 94: Use the classnames library for add classes with condition

src/App.tsx
```diff
@@ -15,37 +17,100 @@ export const goodsFromServer = [
...
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {showResetButton && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              dangerHandler();
+            }}
```
volodymyr-soltys97, line 106:
```suggestion
            onClick={dangerHandler}
```

src/App.tsx
```diff
@@ -15,37 +17,100 @@ export const goodsFromServer = [
...
+}
+
+export const App = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.None);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const visibleGoods = getPreparedGoods(goodsFromServer, sortField, isReversed);
+  const showResetButton = sortField || isReversed;
+
+  function dangerHandler() {
```
volodymyr-soltys97, line 64:
```suggestion
  function resetSort() {
```

volodymyr-soltys97 (changes requested):
Almost done!

lerastarynets (approved):
You're on the right track!

PR #2020, opened 2024-08-18
task solution

olshum8:
[DEMO LINK](https://olshum8.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -16,35 +17,98 @@ export const goodsFromServer = [
...
+
+  const visibleGoods = [...goodsFromServer].sort(
+    (good1: string, good2: string) => {
+      switch (sortField) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+        case SortType.RESET:
+          return 0;
```
VitaliyBondarenko1982, line 37:
```suggestion
```
redundant.
and better to sort inside SortType.ALPHABET and SortType.LENGTH cases. it will avoid redundant iteration when no need to sort goods.

suggest to make function outside of compenent which will except goods from server, sortField and isVisible state and return visibleGoods. inside component just call this function.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check demo link - now it is render readme.md file (ask in chat if will have questions with this problem)

Check comments.

PR #2022, opened 2024-08-18
add solution

Mazurok001:
[DEMO LINK](https://Mazurok001.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


Teollan (approved):
Brilliant 💎 !

src/App.tsx
```diff
@@ -15,36 +16,112 @@
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {sortGoods.map(good => (
+          <li data-cy="Good" key={good}>
```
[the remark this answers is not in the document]
Mazurok001: ок, I add index for goods name.

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortBy {
+  Alphabetically = 'alphabetically',
+  Length = 'length',
+}
+
+interface SortParams {
+  sortField: string;
```
BudnikOleksii, line 25: It's not a string, use your enum `SortBy`

src/App.tsx
```diff
@@ -15,36 +16,112 @@
...
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={classNames('button', 'is-warning', {
+            'is-light': !isReversed,
+          })}
+          onClick={() => {
```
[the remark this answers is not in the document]
BudnikOleksii: > [CODE STYLE] - The function assigned to the `onClick` for the 'Reverse' button should be named to reflect its action. For example, `handleReverse`.

Good point to create a handlers for events

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
+            setIsReversed(!isReversed);
```
BudnikOleksii, line 99: When you need to change state depending on the previous one use functional `setState`

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {sortGoods.map(good => (
+          <li data-cy="Good" key={good + sortGoods.indexOf(good)}>
```
BudnikOleksii, line 121:
```suggestion
          <li data-cy="Good" key={good}>
```
- You can get the index in the iteration method itself as an argument, so `indexOf` doesn't make sense
- It's a bad idea to generate keys at the same time when you render them, it's the same as usage index as a key. In this case good itself as a key is enough

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
+  }
+
+  return preparedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState('');
+  const [isReversed, setIsReversed] = useState(false);
+
+  const sortGoods = getPreparedGoods(goodsFromServer, {
```
```suggestion
  const sortedGoods = getPreparedGoods(goodsFromServer, {
```
`sortGoods` looks like a function name, as it starts from a verb

BudnikOleksii (changes requested):
Good work 👍 

PR #2050, opened 2024-09-17
Enum added

CamilleHaus:
[DEMO LINK](https://CamilleHaus.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
       },
       "devDependencies": {
         "@cypress/react18": "^2.0.1",
-        "@mate-academy/scripts": "^1.8.5",
+        "@mate-academy/scripts": "^1.9.12",
         "@mate-academy/students-ts-config": "*",
         "@mate-academy/stylelint-config": "*",
         "@types/node": "^20.14.10",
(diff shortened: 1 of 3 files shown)


pedro-ruas (changes requested):
Camille, it works and is ok, but I believe that you could use another hook here, `useEffect`, and create an effect based on `sortType`.
Then, when `sortType` changes, the effect could trigger the needed sorting. I believe this is a good exercise to practice

CamilleHaus:
> Camille, it works and is ok, but I believe that you could use another hook here, `useEffect`, and create an effect based on `sortType`. Then, when `sortType` changes, the effect could trigger the needed sorting. I believe this is a good exercise to practice

@pedro-ruas entendi. é que eu nunca usei o useEffect para quando não é uma requisição de API, não tinha pensado nisso.

jv-aquino (approved):
Nice job Camille! About useEffect: the most common to use it is in api and async requests, you're right! But we can also use it to listen to var changes which would be basically a way to simplify the creation of events in js

CamilleHaus:
> Nice job Camille! About useEffect: the most common to use it is in api and async requests, you're right! But we can also use it to listen to var changes which would be basically a way to simplify the creation of events in js

Entendi @jv-aquino, obrigada!

PR #2067, opened 2024-09-29
add solution

YuriiMykhailenko:
🥹
[DEMO LINK](https://YuriiMykhailenko.github.io/react_list-of-goods-hooks/)


src/Components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,12 @@
+import { ListItem } from './ListItem';
+import { ListProps } from '../../types/list/ListProps';
+
+export const GoodsList: React.FC<ListProps> = ({ goods }) => {
+  return (
+    <ul>
+      {goods.map(good => (
+        <ListItem good={good} key={good} />
```
hma-3: I don't think you need a separate component for this

hma-3 (approved):
LGTM!
But there are some possible improvements:
1. You can store enums and types in the same folder
2. I don't think you need to create a separate component for each button (if you want to do is, make it reusable - for example one component for every Sorting button)
3. Your interfaces are unique, so you can probably just put them in the same file as the component they are used with

PR #2068, opened 2024-09-29
solution

Svidruk:
[DEMO LINK](https://Svidruk.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,35 +16,108 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortOption {
+  default = 'default',
```
YuriiMykhailenko, line 20: You don't need to make it unless you use it somewhere
```suggestion
  DEFAULT,
```

YuriiMykhailenko (approved):
Looks good ✨ Just consider my comment below (and also you could try to decompose it a little, just for a practice).

PR #2071, opened 2024-09-30
add task solution

dokrod:
[DEMO LINK](https://dokrod.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -14,37 +15,110 @@ export const goodsFromServer = [
...
+  if (sortField) {
+    preparedGoods = preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.BY_ALPH:
+          return good1.localeCompare(good2);
+
+        case SortType.BY_LENGTH:
+          return good1.length - good2.length;
+
+        default:
```
KrivorotkoYurii, line 44:
```suggestion
        case SortType.DEFAULT:
        default:
          return 0;
```

src/App.tsx
```diff
@@ -14,37 +15,110 @@ export const goodsFromServer = [
...
   'Garlic',
 ];
+enum SortType {
+  BY_ALPH = 'alph',
+  BY_LENGTH = 'length',
+  DEFAULT = '',
+}
+
+type FilterProps = {
+  sortField: SortType;
```
KrivorotkoYurii, line 25: Here i dont realy get why you used a 'SortType' type to the sortField, if sortField is always has a string type

src/App.tsx
KrivorotkoYurii: You can also add a default case

KrivorotkoYurii (approved):
Goooood job!))🔥🔥🔥

PR #2073, opened 2024-09-30
Solution

stasLiutyww:
[DEMO LINK](https://Sukulynt.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-import React, { useState } from 'react';
-import cn from 'classnames'
+import React, { useState, useMemo } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -24,9 +24,9 @@ enum SortType {
 
 function prepareGoods(
   goods: string[],
-  sortedField: string,
+  sortedField: SortType,
   isReversed: boolean,
-) {
+): string[] {
   const copyGoods = [...goods];
 
   if (sortedField !== SortType.default) {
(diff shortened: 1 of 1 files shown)


EustasTheMonk (approved):
Looks crazy mega good!

src/App.tsx
```diff
@@ -15,36 +16,100 @@ export const goodsFromServer = [
...
+
+  return copyGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortedField, setSortField] = useState(SortType.default);
+  const [isReversed, setIsReversed] = useState(false);
+  const visibleGoods = prepareGoods(goodsFromServer, sortedField, isReversed);
+
+  function handleReset() {
```
polosanya, line 59: It is better to use the arrow function to avoid issues with scope as App component is an arrow function

src/App.tsx
```diff
@@ -15,36 +16,100 @@ export const goodsFromServer = [
...
+          className={cn('button is-warning', { 'is-light': !isReversed })}
+          onClick={() => setIsReversed(previousValue => !previousValue)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortedField !== '' || isReversed) && (
```
polosanya: Would be great to create a variable for this condition for better readability

src/App.tsx
```diff
@@ -15,36 +16,100 @@ export const goodsFromServer = [
...
+      switch (sortedField) {
+        case SortType.alphabet:
+          return elementA.localeCompare(elementB);
+
+        case SortType.length:
+          return elementA.length - elementB.length;
+
+        case SortType.default:
+        default:
+          return 0;
```
polosanya: Pay attention that for the default case sort method would be called without any sorting, which is not good. 
Consider not calling sort method at all if it's the default case

polosanya (changes requested):

BudnikOleksii (approved):
Good job 👍

PR #2074, opened 2024-09-30
Solution

vikshiyan:
[DEMO LINK](https://vikshiyan.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


PikalovAndrey (approved):
LGTM! Only one suggestion from me is to move your goodsFromServer to another file, also I think you can do the same for enum and Props.

src/App.tsx
```diff
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
```
VolodymyrKirichenko, line 6: move it to another directory and file - `utils/constants.ts`

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortField {
```
VolodymyrKirichenko, line 19: move it to another directory and file - models/enums/sortField.ts

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
+  LENGTH = 'length',
+  DEFAULT = '',
+}
+
+type Props = {
+  sortField: string;
+  isReverse: boolean;
+};
+
+function getPreperedGoods(goods: string[], { sortField, isReverse }: Props) {
```
VolodymyrKirichenko, line 30: move it to `utils` directory

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortField {
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+  DEFAULT = '',
+}
+
+type Props = {
```
VolodymyrKirichenko, line 25: props???
Props - only for components! do not use this name for other types to avoid confusion in the future
```suggestion
type ...... = { // SortParams or SortConfig will be good i think
```

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn('button is-warning', {
+            'is-light': !isReverse,
+          })}
+          onClick={() => setReverse(value => !value)}
```
VolodymyrKirichenko, line 91: don't forget about handle functions)

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortField(SortField.DEFAULT);
+              setReverse(false);
+            }}
+          >
+            Reset
+          </button>
+        )}
```
VolodymyrKirichenko, line 107: 1) In React, writing sortField !== '' is equivalent to just sortField because non-empty strings evaluate as true in a boolean context, while an empty string evaluates as false.
2) create handleClick function
```suggestion
        {(sortField || isReverse) && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={handleClick}
          >
            Reset
          </button>
        )}
```

VolodymyrKirichenko (changes requested):
Looks good, but let's make some improvements 🥹

PR #2077, opened 2024-10-01
Add task solution

WatarJoy:
https://WatarJoy.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+}
+
+const preparedGoods = [...goodsFromServer];
+
+const sortGoods = (order: SortType, isReversed: boolean): string[] => {
+  let sortedGoods: string[];
+
+  switch (order) {
```
Ostkreuzzz: u can use switch inside of sorting function and sort only one time. for default case just return 0;

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+    setSortOrder(order);
+  };
+
+  const reverseGoods = () => {
+    setGoods(prevGoods => [...prevGoods].reverse());
+    setIsReversed(prevState => !prevState);
+  };
+
+  const resetGoods = () => {
+    setGoods(preparedGoods);
```
Ostkreuzzz: u don't need to set data as you will return a default value anyway

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    sortedGoods.reverse();
+  }
+
+  return sortedGoods;
+};
+
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState<string[]>(preparedGoods);
```
Ostkreuzzz: you don't need another state fo the data here it's redundant. Your page reloads based on click, not data

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+  const [sortOrder, setSortOrder] = useState<SortType>(SortType.DEFAULT);
+
+  const handleSort = (order: SortType) => {
```
Ostkreuzzz: you can get rid of this construction if you will not use state for data

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+
+  const isResetVisible = sortOrder !== SortType.DEFAULT || isReversed;
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${sortOrder !== SortType.ALPHABET ? 'is-light' : ''}`}
```
Ostkreuzzz: classnames will be a better option (+ 'is-' classes will conflic with eachoter )

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+          onClick={() => handleSort(SortType.ALPHABET)}
+        >
           Sort alphabetically
```
Ostkreuzzz: if you will pass that namings to enum, u can use them here as values
`enum SortType {
  ABC = 'Sort alphabetically',
  LENGTH = 'Sort by length',
  RESET = 'Reset',
}`

Ostkreuzzz (approved):
LGTM, but better check for notes, mentor may reject your PR

PR #2078, opened 2024-10-01
Solution

Hradivnyk:
[DEMO LINK](https://Sholudyvyy.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,6 +3,8 @@ import 'bulma/css/bulma.css';
 import './App.scss';
 import { useState } from 'react';
 import classNames from 'classnames';
+import { SortType } from './types/SortType';
+import { sortGoods } from './utils/sortGoods';
 
 export const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,36 +17,109 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+const SORT_FIELD_ALPHABET: string = 'alphabet';
```
Valeraom: It's better to use enum to store sort types

src/App.tsx
```diff
@@ -15,36 +17,109 @@ export const goodsFromServer = [
...
+const SORT_FIELD_LENGTH: string = 'length';
+const SORT_FIELD_RESET: string = '';
+
+type Props = {
```
Valeraom: As for me, this type is useless because `App` doesn't get any props

Valeraom (approved):
Good job!
There are several corrections🙃

PR #2079, opened 2024-10-01
develop

Ostkreuzzz:
https://ostkreuzzz.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +16,113 @@ export const goodsFromServer = [
...
+    sortedData.reverse();
+  }
+
+  return sortedData;
+};
+
 export const App: React.FC = () => {
+  const [selectedSortType, setSelectedSortType] = useState<SortType | null>(
+    null,
+  );
```
WatarJoy, line 57: Better to use SortType.RESET instead of null

WatarJoy (approved):
Good Job!

PR #2086, opened 2024-10-03
add task solution

D-Ptashevskyi:
https://D-Ptashevskyi.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import cn from 'classnames';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
 
+enum SortType {
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
+  DEFAULT = 'default',
+}
+
+const handleSort = (
+  type: SortType,
+  setGoods: React.Dispatch<React.SetStateAction<string[]>>,
```
BudnikOleksii, line 27: - The utility function shouldn't know anything about the state, just gets goods, sorts and returns them
- utils should be a pure functions https://www.geeksforgeeks.org/pure-functions-in-javascript/

src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
+
+  switch (type) {
+    case SortType.ALPHABET:
+      sortedGoods.sort();
+      break;
+    case SortType.LENGTH:
+      sortedGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      sortedGoods = [...goodsFromServer];
```
BudnikOleksii, line 40: You have already created a copy, so you don't need to do it one more time

src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
+  }
+
+  if (isReversed) {
+    sortedGoods.reverse();
+  }
+
+  setGoods(sortedGoods);
+};
+
+const handleReverse = (
```
BudnikOleksii, line 50: - You reverse goods in `handleSort`, so you don't need this one
- You don't need update the goods state. It could be just a variable

src/App.tsx
```diff
@@ -15,36 +16,116 @@ export const goodsFromServer = [
...
+};
+
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState<string[]>(goodsFromServer);
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+  const [sortType, setSortType] = useState<SortType>(SortType.DEFAULT);
+
+  const shouldShowResetButton =
+    (sortType !== SortType.DEFAULT || isReversed) &&
+    JSON.stringify(goods) !== JSON.stringify(goodsFromServer);
```
BudnikOleksii, line 65:
```suggestion
  const preparedGoods = handleSort(goods, { isReversed, sortType });
  const isResetButtonShown = (sortType !== SortType.DEFAULT || isReversed);
```

BudnikOleksii (changes requested):
Good work 👍 
Mostly the same comments that I've left in the similar task
https://github.com/mate-academy/react_list-of-goods-js/pull/1105

src/App.tsx
igorkruz: Not fixed!

src/App.tsx
igorkruz: Not fixed!

src/App.tsx
igorkruz: Not fixed!

src/App.tsx
igorkruz: Not fixed!

igorkruz (changes requested):
Good Job 👍 
Fix all suggestions

djtyrf312 (changes requested):
not a qa task

PR #2106, opened 2024-10-28
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,51 +1,70 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { useState } from 'react';
+import goodsFromServer from './db/goods.json';
+import { SortType } from './utils/enums/sortType';
+import GoodsList from './components/GoodsList';
+import sortGoods from './utils/sortGoods/sortGoods';
 
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+export const App = () => {
+  const [sortType, setSortType] = useState<SortType>(SortType.SORT_BY_DEFAULT);
+  const [isReverse, setReverse] = useState(false);
 
-export const App: React.FC = () => {
+  const goods = sortGoods(goodsFromServer, sortType, isReverse);
+
+  const handleReset = () => {
+    setSortType(SortType.SORT_BY_DEFAULT);
+    setReverse(false);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={cn('button is-info', {
+            'is-light': sortType !== SortType.SORT_BY_CHAR,
+          })}
+          onClick={() => setSortType(SortType.SORT_BY_CHAR)}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={cn('button is-success', {
+            'is-light': sortType !== SortType.SORT_BY_LENGTH,
+          })}
+          onClick={() => setSortType(SortType.SORT_BY_LENGTH)}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn('button is-warning', {
+            'is-light': !isReverse,
+          })}
+          onClick={() => {
+            setReverse(!isReverse);
+          }}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortType.length || isReverse) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={handleReset}
+          >
+            Reset
+          </button>
+        )}
       </div>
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
```
VolodymyrKirichenko: move ur `enum` to a separate file - `utils/enums/sortType.ts`

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+  SORT_BY_LENGTH = 'SORT_BY_LENGTH',
+}
+
+type TGoods = string[];
+
+interface IGoodsProps {
+  goods: Readonly<TGoods>;
+}
+
+const Goods: FC<IGoodsProps> = ({ goods }) => (
```
VolodymyrKirichenko: move it to a separate component = `GoodsList.tsx`

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+  <ul>
+    {goods.map(el => (
+      <li data-cy="Good" key={el}>
+        {el}
+      </li>
+    ))}
+  </ul>
+);
+
+type TSortGoods = (
```
VolodymyrKirichenko: why did u use `interface` above, and `type` here?
Never mix style ur code. Pick something 1 and use it

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+  goods: TGoods,
+  sortType: typeof SortType | string,
+  isReverse: boolean,
+) => TGoods;
+
+const sortGoods: TSortGoods = (goods, sortType, isReverse = false) => {
```
VolodymyrKirichenko: move it ti another directory - `utils/sortGoods/sortGoods.ts`

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+    }
+  });
+
+  return isReverse && sortType === SortType.SORT_BY_LENGTH
+    ? sortedGoods.reverse()
+    : sortedGoods;
+};
+
+export const App = () => {
+  const [sortType, setSortType] = useState('');
```
VolodymyrKirichenko: 1) use ur `enum` in generic type
```suggestion
  const [sortType, setSortType] = useState<SortType>(SortType.Default);
```
2) Add default sorting
```
enum SortType {
  SORT_BY_Default = '',
  SORT_BY_CHAR = 'SORT_BY_CHAR',
  SORT_BY_LENGTH = 'SORT_BY_LENGTH',
}
```

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+            setReverse(!isReverse);
+          }}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortType !== '' || isReverse) && (
```
VolodymyrKirichenko: will be the same, but syntax is better
```suggestion
        {(sortType.length || isReverse) && (
```

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortType('');
+              setReverse(false);
+            }}
```
VolodymyrKirichenko: create - `handleReset` function for it

src/App.tsx
```diff
@@ -15,37 +16,110 @@ export const goodsFromServer = [
...
+      <li data-cy="Good" key={el}>
+        {el}
+      </li>
+    ))}
+  </ul>
+);
+
+type TSortGoods = (
+  goods: TGoods,
+  sortType: typeof SortType | string,
```
VolodymyrKirichenko: use `enum` only

```suggestion
  sortType: SortType,
```

VolodymyrKirichenko (changes requested):
Looks good, but u can do it better 👍 
Let's improve ur code a bit

src/App.tsx
ab3MN: I think it would be better to create a type catalog and keep all types (types, interfaces, enums, etc.) to be sorted in a SortTypes file, because then it would be easy to search for the types you need in one file named after the feature it is used in.

src/App.tsx
ab3MN: I think there is no difference and for me using types for functions is more understandable and readable

src/App.tsx
ab3MN: I think there is no difference and for me using types for functions is more understandable and readable

src/App.tsx
VolodymyrKirichenko: `enum` is a type that can be used in dozens of places in a project. NOT only in 1 function. Therefore, they are placed in a separate directory
`enum` - not bound to 1 function

src/App.tsx
VolodymyrKirichenko: in ur case, the `interface` or `type` doesn't give u any bonuses, so I don't see the point in mixing. 
In my opinion, on the contrary, it is not readable, cuz ur code is mixed

If u want to use `interface` for components and `type` for everything else, then do so in all cases in the future.

but I wouldn't do that) 😄

VolodymyrKirichenko (approved):
Well done 🔥

PR #2120, opened 2024-11-04
Develop

Krykunov:
[DEMO LINK](https://Krykunov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
 
+enum SortField {
+  name = 'name',
+  length = 'length',
+}
+
+interface FilterParams {
+  field: SortField | '';
+  isReverse: boolean;
+}
```
ab3MN, line 28: u can add default sorting to enum and use it and move it to another directory
```
enum SortField {
  name = 'name',
  length = 'length',
  default = ''
}
```

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+
+const prepareGoods = (goods: string[], { field, isReverse }: FilterParams) => {
+  const preparedGoods = [...goods];
```
ab3MN, line 31: it would be better to keep it into utils

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
+          return 0;
+      }
+    });
+  }
+
+  return isReverse ? preparedGoods.reverse() : preparedGoods;
+};
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortField | ''>('');
```
ab3MN, line 50: u can use default sort field in enum

src/App.tsx
```diff
@@ -15,36 +17,98 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {showGoods.map(good => (
+          <li key={good} data-cy="Good">
+            {good}
+          </li>
+        ))}
       </ul>
     </div>
```
ab3MN, line 113: it  would be better to move it to a separate component like GoodsList

ab3MN (approved):
Good job👍

src/App.tsx
BudnikOleksii: > u can use default sort field in enum

Good point

src/App.tsx
BudnikOleksii: > it would be better to move it to a separate component like GoodsList

Good point

BudnikOleksii (approved):
Good job 👍

PR #2172, opened 2024-12-03
Solution

andrushchenkoo:
[DEMO LINK](https://andrushchenkoo.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
```
andrushchenkoo, line 4: haven't transferred the edits from the list-of-goods task yet

Welbrn (approved):
Good job!

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+
+  switch (sortType) {
+    case SortType.Alphabet:
+      sortedGoods = sortedGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.Length:
+      sortedGoods = sortedGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      sortedGoods = [...goods];
```
BudnikOleksii, line 39:
```suggestion
    case SortType.Alphabet:
      sortedGoods.sort((a, b) => a.localeCompare(b));
      break;
    case SortType.Length:
      sortedGoods.sort((a, b) => a.length - b.length);
      break;
    default:
      break;
```
`sort` mutates an array, so you don't need to assign a new value for the variable. And you don't need to create a copy one more time

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+    setIsReverse(false);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${activeSort === 'alphabet' ? '' : 'is-light'}`}
```
BudnikOleksii, line 77: To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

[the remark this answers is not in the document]
BudnikOleksii: > Consider using the `SortType.Alphabet` enum value instead of the string 'alphabet' for consistency and to avoid potential errors if the enum value changes.

Good point

src/App.tsx
```diff
@@ -15,36 +15,104 @@
...
+          onClick={sortAlphabetically}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={`button is-success ${activeSort === 'length' ? '' : 'is-light'}`}
```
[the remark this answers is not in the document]
BudnikOleksii: > Consider using the `SortType.Length` enum value instead of the string 'length' for consistency and to avoid potential errors if the enum value changes.

Fix in all places

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+          onClick={sortByLength}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-warning ${isReverse === true ? '' : 'is-light'}`}
```
BudnikOleksii, line 93: You don't need to compare with `true`, it's already boolean

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
+          onClick={goodsReverse}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(activeSort !== '' || isReverse) && (
```
BudnikOleksii, line 99: You don't need to compare with an empty string, just check if a value exists
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/App.tsx
```diff
@@ -15,36 +15,104 @@ export const goodsFromServer = [
...
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goods.map((good, index) => (
+          <li data-cy="Good" key={goods[index]}>
```
BudnikOleksii, line 112:
```suggestion
          <li data-cy="Good" key={good}>
```
You don't need to access with index as you have an element here

BudnikOleksii (changes requested):
Good work 👍 

PR #2175, opened 2024-12-03
react+TS

AsyaDev14:
Demo : https://AsyaDev14.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState, FC } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


andrushchenkoo (approved):
The code looks good, all the requirements of the check sheet are fulfilled

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+export enum SortKeys {
+  ALPHABETICAL = 'Alphabetical',
+  LENGTH = 'Length',
```
BudnikOleksii, line 20:
```suggestion
export enum SortType {
  Alphabetical = 'alphabetical',
  Length = 'length',
```
Enum properties shouldn't be in upper case, check the example from the [doc](https://mate-academy.github.io/fe-program/js/extra/typescript)
image

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+}
+
+export const App: FC = () => {
+  const [sort, setSort] = useState<string>('');
```
```suggestion
  const [sort, setSort] = useState<SortType>(SortType.Default);
```
It's not a string, use corresponding type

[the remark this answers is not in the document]
BudnikOleksii: > The initial state of `sort` is set to an empty string. Consider using a default value from the `SortKeys` enum to maintain consistency and avoid potential issues with type safety or logic errors.

Good point

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+
+  const getSortedList = (): string[] => {
```
BudnikOleksii, line 27: Move this function outside the component

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+    let sortedGoods = [...goodsFromServer];
```
BudnikOleksii, line 28:
```suggestion
    const sortedGoods = [...goodsFromServer];
```
Prefer const over let

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+
+    switch (sort) {
+      case SortKeys.ALPHABETICAL:
+        sortedGoods.sort();
+        break;
+      case SortKeys.LENGTH:
+        sortedGoods.sort((a, b) => a.length - b.length);
+        break;
+      default:
+        sortedGoods = [...goodsFromServer];
```
BudnikOleksii, line 38: You have already created a copy, you don't need to do it one more time

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+    }
+
+    return isReversed ? sortedGoods.reverse() : sortedGoods;
+  };
+
+  const toggleReverse = (): void => {
+    setIsReversed(prev => !prev);
+  };
+
+  const isOriginalOrder = sort === '' && !isReversed;
```
BudnikOleksii, line 48:
```suggestion
  const isOriginalOrder = !sort && !isReversed;
```
You don't need to compare with an empty string, just check if a value exists or not

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${sort === SortKeys.ALPHABETICAL ? '' : 'is-light'}`}
```
BudnikOleksii, line 55: To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

src/App.tsx
```diff
@@ -15,36 +15,85 @@ export const goodsFromServer = [
...
+          onClick={(): void => setSort(SortKeys.ALPHABETICAL)}
+        >
           Sort alphabetically
         </button>
 
-        <button type="button" className="button is-success is-light">
+        <button
+          type="button"
+          className={`button is-success ${sort === SortKeys.LENGTH ? '' : 'is-light'}`}
+          onClick={(): void => setSort(SortKeys.LENGTH)}
```
BudnikOleksii, line 64:
```suggestion
          onClick={() => setSort(SortKeys.LENGTH)}
```
You don't need to specify a type every time, in this case typescript understands

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

Moroz-Dmytro (changes requested):
Fix comments above

PR #2176, opened 2024-12-03
solution

Yevhen-Filimonov:
[DEMO LINK](https://yevhen-filimonov.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goodsFromServer = [
+export const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


schyrva (approved):
Good Job! 👍

src/App.tsx
```diff
@@ -15,36 +15,114 @@ export const goodsFromServer = [
...
+
+      return isReversed ? sortedGoods.reverse() : sortedGoods;
+    });
+    setActiveSort(SortType.Length);
+  };
+
+  const reverseGoods = () => {
+    setGoods(prevGoods => [...prevGoods].reverse());
+    setIsReversed(prevReversed => !prevReversed);
+  };
```
BudnikOleksii, line 60: It isn't easy to understand what is happening here. Why in sort method you reverse and so on?

Actually, you don't need to store goods in a state at all. Store just in a variable as it's a dynamic value depending on the sort type and isReversed. Create a function that prepares goods and use it:
```typescript
const preparedGoods = prepareGoods(goods, { sortType, isReversed });
```

src/App.tsx
```diff
@@ -15,36 +15,114 @@ export const goodsFromServer = [
...
+
+  if (activeSort === SortType.Length) {
+    lengthButtonClass = 'button is-success';
+  }
+
+  let reverseButtonClass = 'button is-warning is-light';
+
+  if (isReversed) {
+    reverseButtonClass = 'button is-warning';
+  }
```
BudnikOleksii, line 78: 🤯 
- Don't create variables for classes in this case.
- You have a lot of duplications
- To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

src/App.tsx
```diff
@@ -15,36 +15,114 @@ export const goodsFromServer = [
...
+
+  const resetButton =
+    goods.join() !== goodsFromServer.join() ? (
```
BudnikOleksii, line 81: That's a bad idea. What if goods will be an array of objects instead of strings?
You can just check if you have any type of sort or `isReversed` instead

src/App.tsx
```diff
@@ -15,36 +15,114 @@ export const goodsFromServer = [
...
+      <button
+        type="button"
+        className="button is-danger is-light"
+        onClick={resetGoods}
+      >
+        Reset
+      </button>
+    ) : null;
```
BudnikOleksii, line 89: You don't need to use the ternary operator if you return nothing in the second case

src/App.tsx
```diff
@@ -15,36 +15,114 @@ export const goodsFromServer = [
...
+    lengthButtonClass = 'button is-success';
+  }
+
+  let reverseButtonClass = 'button is-warning is-light';
+
+  if (isReversed) {
+    reverseButtonClass = 'button is-warning';
+  }
+
+  const resetButton =
```
BudnikOleksii, line 80: And don't create a variable for UI element, do it just in return

BudnikOleksii (changes requested):
Good work 👍 
Let's try to simplify your solution

PR #2178, opened 2024-12-03
Solution

schyrva:
[DEMO LINK](https://schyrva.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,10 +24,10 @@ enum SortType {
 
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<string[]>([...goodsFromServer]);
-  const [sortOrder, setSortOrder] = useState<SortType>(SortType.Default);
+  const [sortField, SetSortField] = useState<SortType>(SortType.Default);
   const [isReversed, setIsReversed] = useState<boolean>(false);
 
-  const applySort = (sortType: SortType) => {
+  const handleSort = (sortType: SortType) => {
     const sortedGoods = [...goodsFromServer];
 
     if (sortType === SortType.Alphabetical) {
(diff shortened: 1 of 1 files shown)


Ilya-Ilchuk (approved):
Good job :+1:

src/App.tsx
```diff
@@ -15,36 +16,87 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  Default = 'default',
+  Alphabetical = 'alphabetical',
+  Length = 'length',
+}
+
 export const App: React.FC = () => {
+  const [goods, setGoods] = useState<string[]>([...goodsFromServer]);
```
BudnikOleksii: You don't need to store goods in state

src/App.tsx
```diff
@@ -15,36 +16,87 @@ export const goodsFromServer = [
...
+  const [sortField, SetSortField] = useState<SortType>(SortType.Default);
```
BudnikOleksii:
```suggestion
  const [sortField, setSortField] = useState<SortType>(SortType.Default);
```

src/App.tsx
```diff
@@ -15,36 +16,87 @@ export const goodsFromServer = [
...
+
+    if (sortType === SortType.Alphabetical) {
+      sortedGoods.sort((a, b) => a.localeCompare(b));
+    } else if (sortType === SortType.Length) {
+      sortedGoods.sort((a, b) => a.length - b.length);
+    }
+
+    setGoods(isReversed ? sortedGoods.reverse() : sortedGoods);
+    SetSortField(sortType);
+  };
```
BudnikOleksii: - You reverse in the sort handler, it's confusing
- Use switch instead of `if else`

To simplify just create a function outside the component that will prepare goods. Store them in a variable and just render them, no need to store them in a state as it's a dynamic value

```typescript
const preparedGoods = prepareGoods(goods, { isReversed, sortType });
```

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

PR #2187, opened 2024-12-04
done task?

Jion353:
[DEMO LINK](https://Jion353.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


LiliiaVol (approved):
Good job!

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+const SOTR_BY_ALPHABET: string = 'alphabet';
+const SOTR_BY_LENGTH: string = 'length';
```
BudnikOleksii, line 19: It's not a string, create corresponding type or enum

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+
+function SortGoodsFrom(
+  originArray: string[],
+  reverse: boolean,
+  sortBy: string,
```
```suggestion
  isReversed: boolean,
  sortBy: sortType,
```
- Boolean variables should start from `is` or `has`
- `sortBy` isn't a string, it should have corresponding type

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+) {
+  let mutateArray: string[] = [...originArray];
+
+  switch (sortBy) {
+    case SOTR_BY_ALPHABET:
+      mutateArray = mutateArray.sort((a, b) => a.localeCompare(b));
```
BudnikOleksii, line 30:
```suggestion
  const preparedGoods = [...originArray];

  switch (sortBy) {
    case SOTR_BY_ALPHABET:
      preparedGoods.sort((a, b) => a.localeCompare(b));
```
- Prefer const over let
- `arr/array/el/elem/obj/object/item/res/result` are bad names, name should describe a variable
- You don't need to assign a new value, `sort` mutates an array

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+  const [isReverse, setIsReverse] = useState(false);
+  const [sort, setSort] = useState('');
+
+  const visibleGoods: string[] = SortGoodsFrom(
+    goodsFromServer,
+    isReverse,
+    sort,
+  );
+
+  function Reset() {
```
BudnikOleksii, line 56: Only classes, constructor functions and components should be in PascalCase.

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+  }
+
+  return mutateArray;
+}
+
 export const App: React.FC = () => {
+  const [isReverse, setIsReverse] = useState(false);
+  const [sort, setSort] = useState('');
+
+  const visibleGoods: string[] = SortGoodsFrom(
```
BudnikOleksii, line 50:
```suggestion
  const visibleGoods = SortGoodsFrom(
```
You don't have to specify type every time, typescript understands what you return from the function

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+    setSort('');
+  }
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-danger ${sort === SOTR_BY_ALPHABET ? null : 'is-light'}`}
```
BudnikOleksii, line 66: To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-danger ${isReverse ? null : 'is-light'}`}
+          onClick={() =>
+            !isReverse ? setIsReverse(true) : setIsReverse(false)
```
BudnikOleksii, line 84:
```suggestion
            setIsReversed(prevState => !prevState)
```
Could be simplified, and when you set state depending on the previous one use functional `setState`

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods.map(goods => (
+          <li data-cy="Good" key={goods}>
+            {goods}
```
BudnikOleksii, line 104:
```suggestion
        {visibleGoods.map(good => (
          <li data-cy="Good" key={good}>
            {good}
```

BudnikOleksii (changes requested):
Good work 👍 
Let's fix some moments and button colors as well, check the working example. In your case they all have the same color
image

PR #2189, opened 2024-12-04
Solution

SemenVodolazskij:
[DEMO LINK](https://SemenVodolazskij.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ export const goodsFromServer = [
 enum SortType {
   ByLength = 'ByLength',
   Alphabetically = 'Alphabetically',
+  Default = '',
 }
 
 function sortGoods(
@@ -43,23 +44,20 @@ function sortGoods(
 }
 
 export const App: React.FC = () => {
-  const [sort, setSort] = useState<SortType | null>(null);
-  const [isReverse, setIsReverse] = useState<boolean>(false);
+  const [sort, setSort] = useState<SortType>(SortType.Default);
+  const [isReversed, setIsReversed] = useState<boolean>(false);
 
-  const isResetButtonVisible = sort || isReverse;
+  const isResetButtonVisible = sort || isReversed;
 
-  const resultSortedGoods =
-    sort || isReverse
-      ? sortGoods(goodsFromServer, sort, isReverse)
-      : goodsFromServer;
+  const resultSortedGoods = sortGoods(goodsFromServer, sort, isReversed);
 
   const handleToggleSortType = (type: SortType) => {
-    setSort(type === sort ? null : type);
+    setSort(type === sort ? SortType.Default : type);
   };
 
   const handleReset = () => {
-    setIsReverse(false);
-    setSort(null);
+    setIsReversed(false);
+    setSort(SortType.Default);
   };
 
   return (
@@ -88,9 +86,9 @@ export const App: React.FC = () => {
         <button
           type="button"
           className={classNames('button', 'is-warning', {
-            'is-light': !isReverse,
+            'is-light': !isReversed,
           })}
-          onClick={() => setIsReverse(!isReverse)}
+          onClick={() => setIsReversed(prev => !prev)}
         >
           Reverse
         </button>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,15 +27,16 @@ function sortGoods(
   reverse: boolean,
 ): string[] {
   const sortedGoods = [...goods].sort((good1, good2) => {
-    if (sortType === SortType.Alphabetically) {
-      return good1.localeCompare(good2);
-    }
+    switch (sortType) {
+      case SortType.Alphabetically:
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -15,36 +16,106 @@ export const goodsFromServer = [
...
+    if (sortType === SortType.Alphabetically) {
+      return good1.localeCompare(good2);
+    }
+
+    if (sortType === SortType.ByLength) {
+      return good1.length - good2.length;
+    }
+
+    return 0;
+  });
```
BudnikOleksii: - You don't need to sort if you don't have any `sortType`
- Use switch case instead

src/App.tsx
```diff
@@ -15,36 +16,106 @@ export const goodsFromServer = [
...
+
+  return reverse ? sortedGoods.reverse() : sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sort, setSort] = useState<SortType | null>(null);
+  const [reverse, setReverse] = useState<boolean>(false);
```
BudnikOleksii: Boolean variables should start from `is` or `has`

src/App.tsx
```diff
@@ -15,36 +16,106 @@ export const goodsFromServer = [
...
+          type="button"
+          className={classNames('button', 'is-info', {
+            'is-light': sort !== SortType.Alphabetically,
+          })}
+          onClick={() => {
+            if (sort === SortType.Alphabetically) {
+              return setSort(null);
+            } else {
+              return setSort(SortType.Alphabetically);
+            }
```
BudnikOleksii: - Actually if you check [working example](https://mate-academy.github.io/react_list-of-goods/) there are no such toggle functionality
- Also you do almost the same for two buttons, if you want to implement a toggle you can create one handler for both.
- Also, you don't need to return anything from this functions
Could be like this:
```typescript
const handleToggleSortType = (type: SortType) => {
  setSort(type === sort ? null : type);
};
```

src/App.tsx
```diff
@@ -15,36 +16,106 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
+        {isResetButtonVisible && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setReverse(false);
+              setSort(null);
+            }}
```
BudnikOleksii: Create a function instead of passing anonim one to the `onClick`

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+        return 0;
+    }
+  });
+
+  return reverse ? sortedGoods.reverse() : sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sort, setSort] = useState<SortType | null>(null);
+  const [isReverse, setIsReverse] = useState<boolean>(false);
```
Moroz-Dmytro:
```suggestion
  const [isReversed, setIsReversed] = useState<boolean>(false);
```

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+      default:
+        return 0;
+    }
+  });
+
+  return reverse ? sortedGoods.reverse() : sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sort, setSort] = useState<SortType | null>(null);
```
Moroz-Dmytro:
```suggestion
  const [sort, setSort] = useState<SortType>(SortType.Default);
```

Create `Default`  option in `enum`

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
+  const [isReverse, setIsReverse] = useState<boolean>(false);
+
+  const isResetButtonVisible = sort || isReverse;
+
+  const resultSortedGoods =
+    sort || isReverse
+      ? sortGoods(goodsFromServer, sort, isReverse)
+      : goodsFromServer;
```
Moroz-Dmytro:
```suggestion
  const resultSortedGoods = sortGoods(goodsFromServer, sort, isReversed);
```

It's not the same?

src/App.tsx
```diff
@@ -15,36 +16,101 @@ export const goodsFromServer = [
...
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={classNames('button', 'is-warning', {
+            'is-light': !isReverse,
+          })}
+          onClick={() => setIsReverse(!isReverse)}
```
Moroz-Dmytro:
```suggestion
          onClick={() => setIsReverse(prev => !prev)}
```

Always use callback when update state based on previous value

Moroz-Dmytro (changes requested):
Good work 👍 
Some small comments below

BudnikOleksii (approved):

PR #2197, opened 2024-12-07
Solution

IlMarauder:
https://IlMarauder.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { useState } from 'react';
+
+import { Goods } from './Good/Good';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,37 +18,94 @@ export const goodsFromServer = [
...
+  if (reverse) {
+    preparedGoods.reverse();
+  }
+
+  return preparedGoods;
+}
+
+export const App = () => {
+  const [sortType, setSortType] = useState(SortType.None);
+  const [reverse, setReverse] = useState(false);
```
BudnikOleksii, line 56: Boolean variables should start from `is` or `has`

src/App.tsx
```diff
@@ -15,37 +18,94 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={cn('button is-warning', {
+            'is-light': !reverse,
+          })}
+          onClick={() => {
+            setReverse(!reverse ? true : false);
```
BudnikOleksii, line 88:
```suggestion
            setReverse(currentState => !currentState);
```
- It's already boolean, you don't need a ternary operator in this case
- When you set state depending on the previous one use functional `setState`

BudnikOleksii (approved):
Good work 👍

PR #2226, opened 2025-01-03
Solution

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+enum ESortType {
+  ALPHABETICALLY = 'alphabetically',
+  LENGTH = 'by length',
+  DEFAULT = '',
+}
+
+interface IOptions {
+  sortBy: ESortType;
+  isReversed: boolean;
+}
```
naziamoff: what `I/E` means?

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum ESortType {
+  ALPHABETICALLY = 'alphabetically',
+  LENGTH = 'by length',
```
naziamoff: usually we use snake case. But I'd make it simpler and just keep `alphabet/length` in both enum keys and values

```suggestion
  LENGTH = 'by_length',
```

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+  DEFAULT = '',
+}
+
+interface IOptions {
+  sortBy: ESortType;
+  isReversed: boolean;
+}
+
+const getGoods = (goods: string[], { sortBy, isReversed }: IOptions) => {
```
naziamoff: We can keep all the stuff in a single props object.
It looks a bit messy when some props types are defined in the interface and some in the function directly

```suggestion
interface Options { // you have it above, added it as an example
  goods: string[];
  sortBy: SortType;
  isReversed: boolean;
}

const getGoods = ({ goods, sortBy, isReversed }: Options) => {
```

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+    visibleGoods = visibleGoods.toReversed();
+  }
+
+  if (sortBy) {
+    switch (sortBy) {
+      case ESortType.ALPHABETICALLY:
+        visibleGoods = visibleGoods.toSorted();
+        if (isReversed) {
+          visibleGoods = visibleGoods.toReversed();
+        }
```
naziamoff: It looks like you already did it on line 34, didn't you? I'm wondering whether it doesn't affect the behavior and reverse the thing twice

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+
+        break;
+      case ESortType.LENGTH:
+        if (isReversed) {
+          visibleGoods = visibleGoods.toSorted((a, b) => b.length - a.length);
+        } else {
+          visibleGoods = visibleGoods.toSorted((a, b) => a.length - b.length);
+        }
```
naziamoff: what if we move reverse logic to the end (after switch) and execute it only there? Will it work as expected? I mean if we already sorted by length - having reversed it we'll get the expected order anyway, no need to sort in reverse direction. But I might be missing sth

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortBy(ESortType.DEFAULT);
+              setIsReversed(false);
+            }}
+          >
+            Reset
+          </button>
+        ) : null}
```
naziamoff: we can use `&&` operator, it will work the same and there will be no need in empty ternary condition
```suggestion
        {canReset && (
          <button
            type="button"
            className="button is-danger is-light"
            onClick={() => {
              setSortBy(ESortType.DEFAULT);
              setIsReversed(false);
            }}
          >
            Reset
          </button>
        )}
```

src/App.tsx
```diff
@@ -15,36 +16,110 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {visibleGoods &&
+          visibleGoods.map(good => (
+            <li key={good} data-cy="Good">
+              {good}
+            </li>
+          ))}
```
naziamoff: You can assign an empty array to visibleGoods variable right away so you won't need to take care of it in the rendering logic

```suggestion
     const visibleGoods = getGoods(props) || []; // if null or undefined it will be empty array
    
      return (
          ...some JSX stuff
         {visibleGoods.map(good => ( // if array is empty, there will be nothing rendered
            <li key={good} data-cy="Good">
              {good}
            </li>
          ))}
```

naziamoff (approved):

src/App.tsx
djtyrf312: I - interface, E - enum. It is useful for importing such types. You can understand what type it is

src/App.tsx
djtyrf312: thanks, every comment was fixed

src/App.tsx
naziamoff: I never faced this indication before tbh and I don't think this is necessary.

I mean if we have variable typed with SortType it is kind of clear that we're dealing with enum by its name. Also the thing with typescript is that it displays errors instantly if you used some type incorrectly. 

So I'd say that it is better to keep names clean because it is almost the same as naming things like `goodsArr` or `shouldRenderListBool`

src/App.tsx
```diff
@@ -15,36 +16,102 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-export const App: React.FC = () => {
+enum SortType {
+  ALPHABETICALLY = 'alphabet',
+  LENGTH = 'length',
+  DEFAULT = '',
```
naziamoff: I would call it `None` explicitly instead of `Default` with empty string. So by reading the code I can understand that no sorting will be applied

src/App.tsx
```diff
@@ -15,36 +16,102 @@ export const goodsFromServer = [
...
+}
+
+interface Options {
+  goods: string[];
+  sortBy: SortType;
+  isReversed: boolean;
+}
+
+const getGoods = ({ goods, sortBy, isReversed }: Options) => {
```
naziamoff: I'd make it `getSortedGoods`  so it is clear there's some action happening

src/App.tsx
```diff
@@ -15,36 +16,102 @@ export const goodsFromServer = [
...
+  return visibleGoods;
+};
+
+export const App = () => {
+  const [sortBy, setSortBy] = useState<SortType>(SortType.DEFAULT);
+  const [isReversed, setIsReversed] = useState(false);
+  const canReset = sortBy || isReversed;
+
+  const visibleGoods =
+    getGoods({ goods: goodsFromServer, sortBy, isReversed }) || [];
```
naziamoff: When all the stuff doesn't fit into one line, we usually format it like this:

```suggestion
  const visibleGoods = getGoods({ 
    goods: goodsFromServer, 
    sortBy, 
    isReversed 
  }) || [];
```

djtyrf312: thanks, fixed the comments above

PR #2257, opened 2025-01-21
add task solution

naz4ik:


src/App.tsx
```diff
@@ -15,36 +16,93 @@ export const goodsFromServer = [
...
 
-export const App: React.FC = () => {
+function sortingGoods(goods: string[], field: string, isReversed: boolean) {
+  const sorted = [...goods];
+
+  if (field === 'Sort alphabetically') {
+    sorted.sort((a, b) => a.localeCompare(b));
+  } else if (field === 'Sort by length') {
+    sorted.sort((a, b) => a.length - b.length);
+  }
```
etojeDenys, line 26: it's better to use switch case in this case

src/App.tsx
```diff
@@ -15,36 +16,93 @@ export const goodsFromServer = [
...
+
+  if (isReversed) {
+    sorted.reverse();
+  }
+
+  return sorted;
+}
+
+export const App = () => {
+  const [goods, setGoods] = useState(goodsFromServer);
```
etojeDenys, line 36: this state looks redundant

etojeDenys:
```suggestion
  const goods = sortingGoods(goodsFromServer, .............);
```

src/App.tsx
```diff
@@ -15,36 +16,93 @@ export const goodsFromServer = [
...
+
+  const reset = () => {
+    setSortField('');
+    setIsReversed(false);
+    setGoods(goodsFromServer);
+  };
+
+  const isOriginalOrder = goods.every(
+    (good, index) => good === goodsFromServer[index],
+  );
```
etojeDenys, line 60: just check if sortField and isReversed specified

etojeDenys (changes requested):

PR #2358, opened 2025-03-13
task solution

maponomarenko:
[DEMO LINK](https://maponomarenko.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,18 +33,16 @@ function getPreparedGoods(
 ): string[] {
   const preparedGoods = [...goods];
 
-  if (sortField !== SortType.DEFAULT) {
-    preparedGoods.sort((good1: string, good2: string) => {
-      switch (sortField) {
-        case SortType.SORT_ALPHABETICALLY:
-          return good1.localeCompare(good2);
-        case SortType.SORT_LENGTH:
-          return good1.length - good2.length;
-        default:
-          return 0;
-      }
-    });
-  }
+  preparedGoods.sort((good1: string, good2: string) => {
+    switch (sortField) {
+      case SortType.SORT_ALPHABETICALLY:
+        return good1.localeCompare(good2);
+      case SortType.SORT_LENGTH:
+        return good1.length - good2.length;
+      default:
+        return 0;
+    }
+  });
 
   if (reversed) {
     preparedGoods.reverse();
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,36 +16,91 @@ export const goodsFromServer = [
...
+  SORT_ALPHABETICALLY = 'alphabetically',
+  SORT_LENGTH = 'length',
+}
+
+type Prop = {
+  sortField: SortType;
+  reversed: boolean;
+};
+
+function getPreparedGoods(goods: string[], { sortField, reversed }: Prop) {
```
daniil-trusov: You forgot to write what type the function returns.

src/App.tsx
```diff
@@ -15,36 +16,91 @@ export const goodsFromServer = [
...
+  const preparedGoods = [...goods];
+
+  if (sortField) {
```
daniil-trusov: This check is useless because enum can't be null or undefined, so it will always be "true". You can delete this check completely or change it to:
`if (sortField !== SortType.DEFAULT)`

src/App.tsx
```diff
@@ -15,36 +16,91 @@ export const goodsFromServer = [
...
+    preparedGoods.sort((good1, good2) => {
```
daniil-trusov: You forgot to write type of good1 and good2. Same for `const visibleGoods` on row 56.

daniil-trusov (changes requested):
Good job, however you forgot to write variable type in few places and have useless if statement in getPreparedGoods function. Please check comments for more information.

daniil-trusov (approved):
Awesome job!

src/App.tsx
```diff
@@ -15,36 +16,92 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  DEFAULT = '',
+  SORT_ALPHABETICALLY = 'alphabetically',
+  SORT_LENGTH = 'length',
+}
+
+type Prop = {
```
alexander-ignatow, line 25: The type name is not descriptive

Please pick a better one. For instance `Filter`, `GoodsFilter`, `FilterOptions`, etc

src/App.tsx
```diff
@@ -15,36 +16,92 @@ export const goodsFromServer = [
...
+    reversed,
+  });
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${sortField === SortType.SORT_ALPHABETICALLY ? '' : 'is-light'}`}
```
alexander-ignatow, line 67: For now, you may not fix it, but in the future, please consider using `classnames`

alexander-ignatow (changes requested):
Good! But some improvement are needed

PR #2361, opened 2025-03-13
Develop

VitaliiHutsalo:
[DEMO LINK](https://VitaliiHutsalo.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +16,96 @@ export const goodsFromServer = [
...
+  DEFAULT = '',
+  SORT_ALPHABETICALLY = 'Sort alphabetically',
+  SORT_BY_LENGTH = 'Sort by length',
+}
+
+function getSortedGoods(
+  goods: string[],
+  action: SortType | '',
+  isReversed: boolean = false,
+) {
```
daniil-trusov, line 29: You forgot to write the type that function returns

src/App.tsx
```diff
@@ -15,36 +16,96 @@ export const goodsFromServer = [
...
+    default:
+  }
+
+  return isReversed ? sortedGoods.reverse() : sortedGoods;
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.DEFAULT);
+  const [reversed, setReversed] = useState(false);
+  const goods = getSortedGoods(goodsFromServer, sortField, reversed);
```
daniil-trusov, line 50: You forgot to write goods type

daniil-trusov (approved):
Nice work, but don't forget to write what type the function returns and make sure all variables have a type as well. Please check the comments for places where you missed this.

src/App.tsx
alexander-ignatow: I would say it is not critical in this case

src/App.tsx
alexander-ignatow: Why? Type of gods is calculated from the function's return type

src/App.tsx
```diff
@@ -15,36 +16,96 @@ export const goodsFromServer = [
...
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={
+            sortField === SortType.SORT_ALPHABETICALLY
+              ? 'button is-info'
+              : 'button is-info is-light'
```
alexander-ignatow, line 60: Consider using `classnames` the next time

alexander-ignatow (approved):
Good, but please see my comments for the future!

PR #2364, opened 2025-03-13
task solution

radykhovska:
[DEMO LINK](https://radykhovska.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -14,36 +16,109 @@ export const goodsFromServer = [
...
+    preparedGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.Alphabet:
+          return good1.localeCompare(good2);
+
+        case SortType.Length:
+          return good1.length - good2.length;
+
+        case SortType.Default:
+          return 0;
```
AnOvramenko, line 45: You should get rid of this case because it's going into a sort even if it's unnecessary.
Just check it in if statement like sortField !== SortType.Default

AnOvramenko (approved):
Good Job! 👍 
I left some comments

DanilWeda (approved):
Good job. 
Approved.
Think about code spliting in future. 
Make clean function and move their outside components. 
Then import and use inside.

Many thanks!

PR #2368, opened 2025-03-13
Done task

Iwantschenko:
[DEMO LINK](https://Iwantschenko.github.io/react_list-of-goods-hooks/)


OksanaKuziv (approved):
Good job!
but in your function have one problem 
   default:
        throw new Error('SortField Error!!');
        
This should never happen if the state is correctly managed.

PR #2374, opened 2025-03-16
add files

glebanty19:
[DEMO LINK](https://glebanty19.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
+import { useState } from 'react';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +17,89 @@ export const goodsFromServer = [
...
+          return a.length - b.length;
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
R-Ohirok: It's better to move this block of code to separate function outside the App

src/App.tsx
```diff
@@ -15,36 +17,89 @@ export const goodsFromServer = [
...
+
+  const isSorted = sortField !== '' || isReversed;
```
R-Ohirok: I am not sure that 'reversed' data can be called 'sorted', so I think it's better to name it 'isChanged'

R-Ohirok (approved):
Good job! I have a few tips that could improve your code.

PR #2399, opened 2025-03-24
solution

kostivkostiv:
[DEMO LINK](https://kostivkostiv.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,14 +22,14 @@ enum SortType {
   default = '',
 }
 
-type FilterGoods = {
+type GoodsFilter = {
   sortField: SortType;
   reversed: boolean;
 };
 
 function getPreparedGoods(
   goods: string[],
-  { sortField, reversed }: FilterGoods,
+  { sortField, reversed }: GoodsFilter,
 ) {
   const prepareGoods = [...goods];
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


VladRoz95 (approved):

src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  alphabetically = 'Sort alphabetically',
+  length = 'Sort by length',
+  default = '',
+}
+
+type FilterGoods = {
```
alexander-ignatow: This type name is not good - it sounds like a function

Please rename to something more descriptive, like `GoodsFilter`, etc

src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
...
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={classNames('button', 'is-info', {
+            'is-light': sortField === SortType.alphabetically,
```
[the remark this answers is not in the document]
alexander-ignatow: Probably - need re-check on live demo

alexander-ignatow (changes requested):
Looks good, but please, fix the demo-link

src/App.tsx
```diff
@@ -15,36 +16,107 @@ export const goodsFromServer = [
...
+  const prepareGoods = [...goods];
+
+  if (sortField) {
+    prepareGoods.sort((good1, good2) => {
+      switch (sortField) {
+        case SortType.alphabetically:
+          return good1.localeCompare(good2);
+
+        case SortType.length:
+          return good1.length - good2.length;
```
alexander-ignatow, line 43: No need to fix it - just a thought
It would be slightly more effective to change the sorting condition according to the `isReversed`. Then you will need only to reverse the array in an initial order

alexander-ignatow (approved):
👍

PR #2432, opened 2025-04-15
add task solution

Uusuff:
-[DEMO LINK](https://Uusuff.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/components/buttons/Buttons.tsx
+++ b/src/components/buttons/Buttons.tsx
@@ -1,5 +1,6 @@
 import 'bulma/css/bulma.css';
 import React from 'react';
+import { SortType } from '../../types/SortType';
 
 interface Props {
   isLight: string;
@@ -9,7 +10,7 @@ interface Props {
   setResetVisible: (value: boolean) => void;
   isReverse: boolean;
   setIsReverse: (value: boolean) => void;
-  handleSort: (field: string) => void;
+  sortGoods: (field: SortType) => void;
 }
 
 export const Buttons: React.FC<Props> = ({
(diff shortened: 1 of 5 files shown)


src/components/buttons/Buttons.tsx
```diff
@@ -0,0 +1,74 @@
...
+    [SortType.Length, 'is-success'],
+  ];
+
+  return (
+    <div className="buttons">
+      {buttons.map(([type, style]) => (
+        <button
+          key={type}
+          type="button"
+          className={`button ${style} ${activeButton !== type ? isLight : ''}`}
```
volodymyr-soltys97, line 37: Use the classnames library for add classes with conditions. fix it everywhere

src/components/buttons/Buttons.tsx
```diff
@@ -0,0 +1,74 @@
...
+          onClick={() => {
+            setActiveButton(type);
+            sortGoods(type);
+            setResetVisible(true);
+          }}
```
volodymyr-soltys97, line 42: Move this logic from jsx to the helper function and use it here

src/components/buttons/Buttons.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {resetVisible && (
+        <button
+          type="button"
+          className="button is-danger is-light"
+          onClick={() => {
+            sortGoods(SortType.Reset);
+            setResetVisible(false);
+            setActiveButton('');
+            setIsReverse(false);
+          }}
```
volodymyr-soltys97, line 67: The same as above

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #2492, opened 2025-05-25
solution

izuwiiii:
[DEMO LINK](https://izuwiiii.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,37 +17,103 @@ export const goodsFromServer = [
...
+function getPreaparedGoods(
+  goods: string[],
+  { sortedField, reversed }: FilterParams,
+) {
+  let preparedGoods = [...goods];
+
+  if (sortedField) {
+    preparedGoods.sort((good1, good2) => {
+      switch (sortedField) {
+        case 'alphabet':
```
lidiia-tsymborovych, line 40: You can also use enum instead of string  inside of the function

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
```
lidiia-tsymborovych, line 2: I guess you shouldn't have done it because of this point in readme.md : DON'T use class components (for this and next tasks).

lidiia-tsymborovych (approved):
Good job!

PR #2504, opened 2025-05-26
Develop

RTSaitama:


src/App.tsx
```diff
@@ -15,35 +18,106 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">...</li>
+          {sortedGoods.map(good => {
+            return (
+              <li key={good.id} data-cy="Good">
+                {good.name}
+              </li>
+            );
+          })}
         </ul>
       </ul>
```
koliiis: I think we should remove second ul, but who knows 🤷‍♀️

koliiis (approved):
Great job!

PR #2555, opened 2025-06-17
task

Leh-02:
[DEMO LINK](https://Leh-02.github.io/react_list-of-goods-hooks/)


etojeDenys (changes requested):
fix the demo link and then request the review again
image

PR #2560, opened 2025-06-20
finished

rekverr:
[DEMO LINK](https://rekverr.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
-          Sort alphabetically
+        <button
+          type="button"
+          className={cn('button', 'is-info', {
+            'is-light': active !== SortBy.ALPHABETICALLY,
+          })}
+          onClick={() => sortByAlphabetically()}
```
TarasHoliuk:
```suggestion
          onClick={sortByAlphabetically}
```
Same everywhere

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
+    setIsReversed(!isReversed);
+  };
+
+  const resetBtn = () => {
+    setActive('');
+    setGoods(goodsFromServer);
+    setIsReversed(false);
+  };
+
+  const isOriginalOrder = () =>
```
TarasHoliuk: It's better to check `active` and `isReversed`. Also it can be just beelean value instead of function

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
 
-export const App: React.FC = () => {
+enum SortBy {
+  ALPHABETICALLY = 'Sort alphabetically',
+  LENGTH = 'Sort by length',
+}
+
+export const App = () => {
+  const [goods, setGoods] = useState(goodsFromServer);
+  const [active, setActive] = useState('');
```
TarasHoliuk: active doesn't reflect value. you can rename it to `sortBy`

src/App.tsx
```diff
@@ -15,36 +16,103 @@ export const goodsFromServer = [
...
 ];
 
-export const App: React.FC = () => {
+enum SortBy {
+  ALPHABETICALLY = 'Sort alphabetically',
+  LENGTH = 'Sort by length',
+}
+
+export const App = () => {
+  const [goods, setGoods] = useState(goodsFromServer);
```
TarasHoliuk: Not critical here, but in such cases, it's better to use another approach: instead of saving sorted/filterd items to state, we can just create a common const variable with the result of sort (you can perform sort in a separate function that accepts isReversed and sortBy and returns a sorted array)

General rule: do not store something in state if it can be computed based on other state values/props

TarasHoliuk (changes requested):
Great job! Added a few comments and one general suggestion

volodymyr-soltys97 (approved):
Great! 🔥

PR #2605, opened 2025-07-24
sol

O1ympus:
[DEMO LINK](https://O1ympus.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +15,110 @@ export const goodsFromServer = [
...
+      preparedGoods.reverse();
+    }
+
+    return preparedGoods;
+  };
+
+  const handleSortAlphabetically = () => {
+    const sorted = getSortedGoods(FIELD_ALPHABETIC, isReversed);
+
+    setGoods(sorted);
```
ArturErmolenko: nope we shouldn't change state in handler
```suggestion
```

src/App.tsx
```diff
@@ -15,36 +15,110 @@ export const goodsFromServer = [
...
+  const handleReset = () => {
+    setGoods(goodsFromServer);
+    setSortBy('');
+    setIsReversed(false);
+  };
+
+  const isInitial =
+    goods.every((item, index) => item === goodsFromServer[index]) &&
+    !isReversed;
+
```
ArturErmolenko, line 50: const x = getSortedGoods(....)

src/App.tsx
```diff
@@ -15,36 +15,110 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
+        {goods.map(good => (
+          <li key={good} data-cy="Good">
+            {good}
+          </li>
```
ArturErmolenko, line 104:
```suggestion
        {x.map(good => (
          <li key={good} data-cy="Good">
            {good}
          </li>
```

src/App.tsx
```diff
@@ -15,36 +15,110 @@ export const goodsFromServer = [
...
+
+  const handleReset = () => {
+    setGoods(goodsFromServer);
+    setSortBy('');
+    setIsReversed(false);
+  };
+
+  const isInitial =
+    goods.every((item, index) => item === goodsFromServer[index]) &&
+    !isReversed;
```
ArturErmolenko:
```suggestion
  const isInitial =
    goods.every((item, index) => item === goodsFromServer[index]) &&
    !isReversed;
```
why we need iterate all array, what if we have 10000000 goods?
can we find better way to understand that good is initial?

the absence of filters can tell us that the array is initial?

src/App.tsx
```diff
@@ -15,36 +15,110 @@ export const goodsFromServer = [
...
+    setGoods(sorted);
+    setSortBy(FIELD_LENGTH);
+  };
+
+  const handleReverse = () => {
+    const reversed = [...goods].reverse();
+
+    setGoods(reversed);
+    setIsReversed(prev => !prev);
+  };
```
ArturErmolenko: handlers shouldn't do such work
they should just change filter state
all work is doing in getSortedGoods outside handlers

```suggestion
  const handleSortByLength = () => {
    setSortBy(FIELD_LENGTH);
  };

  const handleReverse = () => {
    setGoods(reversed);
  };
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):
👍

PR #2617, opened 2025-07-31
develop solution

nadiaaamf:
Reimplement the **React list of goods** implemented before using hooks and typescript.


nadiaaamf:
> Reimplement the **React list of goods** implemented before using hooks and typescript.

(https://nadiaaamf.github.io/react_list-of-goods-hooks/)

PR #2625, opened 2025-08-08
list of goods TS solution

NicssLucik02:


src/App.tsx
```diff
@@ -16,34 +16,95 @@ export const goodsFromServer = [
...
+  const [activeSort, setActiveSort] = useState<string>('all');
+  const [isReversed, setIsReversed] = useState<boolean>(false);
+
+  enum SortStatus {
+    All = 'all',
+    Alphabetical = 'alphabetical',
+    Length = 'length',
+  }
+
+  function getSort(): string[] {
```
alexander-ignatow: Move this function outside of the component

```suggestion
  function getSort(Goods: string[], activeSort: SortStatus, isReserved: boolean): string[] {
```

src/App.tsx
```diff
@@ -16,34 +16,95 @@ export const goodsFromServer = [
...
+    const sorted = [...goodsFromServer];
+
+    if (activeSort === SortStatus.Alphabetical) {
+      sorted.sort((a, b) => a.localeCompare(b));
+    } else if (activeSort === SortStatus.Length) {
+      sorted.sort((a, b) => a.length - b.length);
+    }
+
```
alexander-ignatow: prefer switch/case over multiple else/if

src/App.tsx
```diff
@@ -16,34 +16,95 @@ export const goodsFromServer = [
...
+    if (isReversed) {
+      sorted.reverse();
+    }
+
+    return sorted;
+  }
+
+  const resetOnClick = () => {
```
alexander-ignatow: mind the naming (usually handlers start from "handle")

```suggestion
  const handleResetClick = () => {
```

src/App.tsx
```diff
@@ -16,34 +16,95 @@ export const goodsFromServer = [
...
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={
+            activeSort === SortStatus.Alphabetical
+              ? 'button is-info'
+              : 'button is-info is-light'
+          }
```
alexander-ignatow, line 71: optional - but it is a perfect case for classnames or clsx libraries

alexander-ignatow:
```suggestion
          className={classnames('button is-info', 'is-light': activeSort === SortStatus.Alphabetical}
```

alexander-ignatow (changes requested):
Good, but please fix some minor issues

alexander-ignatow (approved):
👍 LGTM

PR #2643, opened 2025-08-13
Develop

DmitriyALEX:
https://DmitriyALEX.github.io/react_list-of-goods-hooks/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,50 +1,97 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
-
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,50 +1,97 @@
...
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+import { goodsFromServer } from './api/goodsFromServer';
+import { SortField } from './types/enams';
 
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortField | ''>('');
```
derpedcatto: It would be better to add a 'None' option to your enum, so your syntax would be a bit more clear. For example:
```ts
enum SortingOptions {
  None,
  Name,
  Length,
}

// ...

const [sortOption, setSortOption] = useState(SortingOptions.None);
```

derpedcatto (approved):
All good.

PR #2649, opened 2025-08-13
task solution

IlchenkoDima:
https://ilchenkodima.github.io/react_list-of-goods-hooks/


YevheniiKa (approved):
Good job, I like your version of solution!

PR #2719, opened 2025-10-06
_{List of Goods TS}_

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { SortType } from './types/SortType';
+import { GoodList } from './components/GoodList';
 
-export const goodsFromServer = [
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -16,36 +18,72 @@ export const goodsFromServer = [
 ];
 
 export const App: React.FC = () => {
+  const [sortType, setSortType] = useState<SortType>(SortType.None);
```
Aiiyuu, line 21: In **TypeScript**, when you use `useState` with an initial value like `SortType.None`, **TypeScript is able to infer** the type of `sortType` automatically based on the initial value. So, you don't always need to specify `<SortType>` if the initial value already makes it clear what the type should be.

```suggestion
  const [sortType, setSortType] = useState(SortType.None);
```

src/App.tsx
```diff
@@ -16,36 +18,72 @@ export const goodsFromServer = [
...
+
+  if (sortType === SortType.Alphabetical) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  } else if (sortType === SortType.Length) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
Aiiyuu: The sorting and reversing logic for `visibleGoods` could be moved into a dedicated function to make things cleaner and more reusable.

You can use the code like here:
```javascript
function getPreparedGoods(goods: string[], sortType: SortType, isReversed: boolean): string[] {
  const sortedGoods = [...goods];

  if (sortType === SortType.Alphabetical) {
    sortedGoods.sort((a, b) => a.localeCompare(b));
  } else if (sortType === SortType.Length) {
    sortedGoods.sort((a, b) => a.length - b.length);
  }

  if (isReversed) {
    sortedGoods.reverse();
  }

  return sortedGoods;
}
```

Usage
```javascript
const visibleGoods = getPreparedGoods(goodsFromServer, sortType, isReversed);
```

Why:
- **Cleaner:** Makes the main code easier to read.
- **Reusable:** You can use getPreparedGoods anywhere.
- **More maintainable:** All the sorting logic is in one place.

Aiiyuu (approved):

Aiiyuu:
> [DEMO LINK](https://S-Khraban.github.io/react_list-of-goods-hooks/)

**Good job!**

PR #2732, opened 2025-10-10
solution

Maevalar:
[DEMO LINK](https://Maevalar.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
+enum SortType {
+  None = '',
+  Alphabetically = 'alphabetically',
+  Length = 'length',
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.None);
+  const [reversed, setReversed] = useState(false);
+  const [isLight, setIsLight] = useState('');
```
Aiiyuu: If you're using **isLight**, then why not name **reversed** as **isReversed**?

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
+
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (reversed) {
+    visibleGoods = visibleGoods.reverse();
+  }
```
Aiiyuu, line 69: You should move it into a dedicated function outside the component. It might be useful for reusability, and your component would look clearer.

src/App.tsx
```diff
@@ -15,36 +16,112 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  None = '',
+  Alphabetically = 'alphabetically',
+  Length = 'length',
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.None);
```
Aiiyuu, line 26: You can remove `<SortType>` because **TypeScript** figures out the **type** from `SortType.None` automatically. This makes the code cleaner and easier to read. No need to write it twice!

```suggestion
export const App: React.FC = () => {
  const [sortField, setSortField] = useState(SortType.None);
```

Aiiyuu (approved):
Good job! but check comments below

PR #2752, opened 2025-10-16
add task solution

JackSavin:
[DEMO LINK](https://jacksavin.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum Sortype {
+  defalut = '',
+  Alphabetically = 'alphabet',
+  Length = 'length',
+}
```
781calmstate: stick to one approach, you either capitalize(first letter should in upper case) every prop, or they all are in lower case. I'd recommend to use the capitalized approach for all properties.

Also you made a typo in the "default" word

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum Sortype {
```
781calmstate: the enum name should be in `PascalCase` - every word starts with letter in upper case

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
...
+
+const getSortedGoods = (
+  goods: string[],
+  sortType: Sortype,
+  isReversed: boolean,
+) => {
+  const sortedGoods = [...goods];
+
+  switch (sortType) {
+    case 'alphabet':
```
781calmstate: use the enum values instead of magic strings, you already specified enum with values, you should use it

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
...
+      sortedGoods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case 'length':
```
781calmstate: same

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
...
+  }
+
+  return sortedGoods;
+};
+
 export const App: React.FC = () => {
+  const [activeSort, setActiveSort] = useState(Sortype.defalut);
+  const [isReversed, setIsReversed] = useState(false);
+
+  const sortedGoods = getSortedGoods(goodsFromServer, activeSort, isReversed);
```
781calmstate, line 57: great decision to move the logic into the helper, it looks very clean

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
...
+  };
+
+  const toggleReverse = () => {
+    setIsReversed(prev => !prev);
+  };
+
+  const resetGoods = () => {
+    setActiveSort(Sortype.defalut);
+    setIsReversed(false);
+  };
```
781calmstate: Use `handle` as a prefix for event handler functions. For example, `handleSortChange`, `handleToggleReverse`, 'handleSortReset'.

more about conventional namings in react you can read [here](https://dev.to/kristiyanvelkov/react-js-naming-convention-lcg)

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
...
+
+  const isModified = activeSort !== '' || isReversed;
```
781calmstate:
```suggestion
  const isModified = activeSort !== SortType.Default || isReversed;
```

781calmstate (changes requested):
Always try to reuse the code you already implemented. You've declared the enum, but instead of using it in your switch case or in `isModified` condition you use just strings.
Imagine all of a sudden you have to change the values of your sorting in code. Instead of changing all in one place by changing the enum's values, you'll have to change the enum's values + look for all those strings in your file/component

src/App.tsx
```diff
@@ -1,4 +1,6 @@
 import React from 'react';
+import { useState } from 'react';
+import cn from 'classNames';
```
781calmstate: `classnames`, not `classNames`.

check the lib name in your package.json in the dependencies :)

src/App.tsx
```diff
@@ -15,36 +17,110 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SorType {
```
781calmstate, line 20: `SortType`, not `SorType` 😄

781calmstate (approved):

PR #2778, opened 2025-11-03
add solution

entodasha:
[DEMO LINK](https://entodasha.github.io/react_list-of-goods-hooks/)


entodasha:
@mateacademy-ai-mentor you ask me to Update the Data Structure: Change goodsFromServer from an array of strings to an array of objects. Each object should have properties like name, price, and age, similar to the phones in the demo.  you are asking me to perform tasks that are not specified in the conditions

brespect (changes requested):
1. Remove the ability to select two options by double clicks 
{4C4336B4-43E8-4409-9E28-B315ACBA4971}

2. Not correctly sorted when selected sort by alphabet after reverse

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
 ];
 
+enum SortType {
+  default = '',
+  SORT_FIELD_ALPHABET = 'alphabet',
+  SORT_FIELD_LENGTH = 'length',
+}
+
 export const App: React.FC = () => {
+  const [sortField, setSortField] = useState<SortType>(SortType.default);
```
Anton-Kuchmasov, line 25:
```suggestion
  const [sortField, setSortField] = useState<SortType | null>(null);
```

Using this, you shouldn't define `default = ''` in your enum

src/App.tsx
```diff
@@ -15,36 +15,95 @@ export const goodsFromServer = [
...
+
+    if (sortField !== SortType.default) {
+      switch (sortField) {
+        case SortType.SORT_FIELD_ALPHABET:
+          goods.sort();
+          break;
+        case SortType.SORT_FIELD_LENGTH:
+          goods.sort((a, b) => a.length - b.length);
+          break;
+      }
```
Anton-Kuchmasov, line 53: don't forget to add `default: break` to switch-case statements

Anton-Kuchmasov (approved):
Good job!
Check out comments, they may be useful in futher tasks

PR #2845, opened 2026-01-13
solution

AdelinaY18:
[DEMO LINK](https://AdelinaY18.github.io/react_list-of-goods-hooks/)


Br0No (approved):
Good job but we need to use enum SortType including the default option

PR #2855, opened 2026-01-13
Solution

torianykhanna:
[DEMO LINK](https://torianykhanna.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,37 +18,53 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+enum SortType {
+  NONE = 'none',
```
de-vasta, line 22: I don't see that there's a reason to set specific string ~~variables~~ values for `SortType`

de-vasta (approved):
Good 👍🏼

PR #2868, opened 2026-01-14
task completed

Staskovych:
[DEMO LINK](https://Staskovych.github.io/react_list-of-goods-hooks/)


src/App.tsx
```diff
@@ -15,35 +15,84 @@ export const goodsFromServer = [
...
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(sortField !== SortType.None || isReversed) && (
+          <button
+            type="button"
+            className="button is-danger"
+            onClick={() => {
```
SenyaKir: It’s better to use handler functions rather than inline logic.

src/App.tsx
```diff
@@ -15,35 +15,84 @@ export const goodsFromServer = [
...
+          onClick={() => setSortField(SortType.Length)}
+        >
           Sort by length
         </button>
 
-        <button type="button" className="button is-warning is-light">
+        <button
+          type="button"
+          className={`button is-warning ${isReversed ? SortType.None : 'is-light'}`}
+          onClick={() => setIsReversed(prev => !prev)}
```
SenyaKir: It’s better to use handler functions rather than inline logic.

SenyaKir (approved):
good job

src/App.tsx
Staskovych: i agree, thanks for the comment, i corrected, and on 70 line too

PR #2871, opened 2026-01-15
react_list-of-goods-hooks

SenyaKir:
[Demo](https://SenyaKir.github.io/react_list-of-goods-hooks/)


Etwaiz (approved):
Nice, but I think the code could be optimized better!

PR #2957, opened 2026-03-25
Add sorting and filtering

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_list-of-goods-hooks/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,41 +22,37 @@ enum SortType {
   byLength = 'length',
 }
 
-enum FilterType {
-  reverse = 'reverse',
-}
-
 export const App: React.FC = () => {
-  const [filterBy, setFilterBy] = useState<FilterType[]>([]);
   const [sortBy, setSortBy] = useState<SortType | ''>('');
+  const [isReversed, setIsReversed] = useState<boolean>(false);
 
   const preparedGoods = [...goodsFromServer];
 
-  if (sortBy === SortType.byAlphabet) {
-    preparedGoods.sort((a, b) => {
-      return a.localeCompare(b);
-    });
+  switch (sortBy) {
+    case SortType.byAlphabet:
+      preparedGoods.sort((a, b) => {
+        return a.localeCompare(b);
+      });
+      break;
+
+    case SortType.byLength:
+      preparedGoods.sort((a, b) => {
+        return a.length - b.length;
+      });
   }
 
-  if (sortBy === SortType.byLength) {
-    preparedGoods.sort((a, b) => {
-      return a.length - b.length;
-    });
-  }
-
-  if (filterBy.includes(FilterType.reverse)) {
+  if (isReversed) {
     preparedGoods.reverse();
   }
 
-  function setFilter(filter: FilterType) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+
+  const preparedGoods = [...goodsFromServer];
+
+  if (sortBy === SortType.byAlphabet) {
+    preparedGoods.sort((a, b) => {
+      return a.localeCompare(b);
+    });
+  }
+
+  if (sortBy === SortType.byLength) {
```
polosanya: Better use switch case here

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+    preparedGoods.sort((a, b) => {
+      return a.length - b.length;
+    });
+  }
+
+  if (filterBy.includes(FilterType.reverse)) {
+    preparedGoods.reverse();
+  }
+
+  function setFilter(filter: FilterType) {
```
polosanya:
```suggestion
  const setFilter = (filter: FilterType) => {
```
It is a good practice to use one way of creating functions per project

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+    setFilterBy(prevFilters => {
+      if (filterBy.includes(filter)) {
+        return prevFilters.filter(i => i !== filter);
+      } else {
+        return [...prevFilters, filter];
```
polosanya: This logic is too complicated. It is better to use boolean for reversed state and just switch it on change

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+          })}
+          onClick={() => setFilter(FilterType.reverse)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(filterBy.length !== 0 || sortBy !== '') && (
```
polosanya: Better create a boolean constant for this state for better reability

src/App.tsx
```diff
@@ -15,36 +17,100 @@ export const goodsFromServer = [
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setFilterBy([]);
+              setSortBy('');
```
polosanya: Let's create a separate function for reset in case you need to change it

polosanya (changes requested):
Please pay attention to the comments

src/App.tsx
```diff
@@ -15,36 +17,93 @@ export const goodsFromServer = [
...
+      preparedGoods.sort((a, b) => {
+        return a.localeCompare(b);
+      });
+      break;
+
+    case SortType.byLength:
+      preparedGoods.sort((a, b) => {
+        return a.length - b.length;
+      });
+  }
```
polosanya, line 42: Don't forget to always add the default case

polosanya (approved):
Good job! Please pay attention to the comment

PR #2961, opened 2026-03-30
Develop

AnastasiiaHombalevska:


brespect (changes requested):
Good progress, but you need to follow next requirement, not in (github/workflows & cypress folders), remove them:
Write the code inside the `src/` folder following the style guides:
[HTML and CSS styleguide](https://mate-academy.github.io/style-guides/htmlcss.html)
[JS styleguide](https://github.com/mate-academy/style-guides/blob/master/javascript.md)

PR #3001, opened 2026-05-05
add task solution

DmitriyS2k:
[DEMO LINK](https://DmitriyS2k.github.io/react_list-of-goods-hooks/)


Tysiachnyk (approved):
Everything works properly! 

I think it would be better moving `getSortedGoods` outside the component and passing `filter` and `reverseGoods` as parameters. This makes it a pure function, avoids closure on component state, and prevents it from being recreated on every render.

src/App.tsx
```diff
@@ -15,36 +16,94 @@ export const goodsFromServer = [
...
+  None = '',
+  Alphabetically = 'alphabetically',
+  ByLength = 'length',
+}
+
+export const App = () => {
+  const [filter, setFilter] = useState<SortType>(SortType.None);
+  const [reverseGoods, setReverseGoods] = useState<boolean>(false);
+
+  function getSortedGoods(goods: string[]): string[] {
```
dothearts, line 29: it's also can be moved outside component. 
In future you can move it in separate `.ts` file

dothearts (approved):

PR #3006, opened 2026-05-05
Solution

SergGun:


src/App.tsx
```diff
@@ -1,6 +1,8 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
+
+type SotrType = 'alpha' | 'length' | null;
```
yana-karpovych: The task requires using `enum SortType` instead of string literals for possible sorting options (including the default one).

yana-karpovych (changes requested):
The task requires using `enum SortType` instead of string literals for possible sorting options (including the default one).

src/App.tsx
SergGun: Fixed

yana-karpovych (approved):
Well done!

src/App.tsx
```diff
@@ -1,8 +1,16 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { useState } from 'react';
 
-export const goodsFromServer = [
+// type SotrType = 'alpha' | 'length' | null;
```
dothearts, line 5: You can remove it

src/App.css
```diff
@@ -0,0 +1,3 @@
+iframe {
+  display: none;
+}/*# sourceMappingURL=App.css.map */
```
dothearts, line 3: You don't need this changes

src/App.css.map
dothearts, line 1: This file doesn't need to be updated

dothearts (approved):
Please, provide DEMO LINK in PR description. 
Also, check comments for future realization. 
Many thanks!

PR #3009, opened 2026-05-06
Develop

rlysunets:
[DEMO LINK](https://rlysunets.github.io/react_list-of-goods-hooks/)


Misharkus (approved):
U need to change classes on buttons. 1st button 'is-info', 3rd button 'is-warning'

dothearts (changes requested):
Many thanks for request. 

Please,  check the commit history: 
Image

In this case you need push only you commits. 
Also, in PR many changes, which not needed in this task. 
Please clean up request. In changes need to see only task neded changes, also commits need to be only yours. 
After that, re-request review.

PR #3011, opened 2026-05-06
Solution

Misharkus:
(https://Misharkus.github.io/react_list-of-goods-hooks/)


Palfinger2002 (approved):
Good job. Approved. But on the line 88  - the condition is correct and works, but for reabability and consistency it's common to use "!isReversed" instead of 
"isReversed === false". Also can check other conditions.

dothearts (approved):

PR #3020, opened 2026-05-08
add solution

Dima2702:
https://Dima2702.github.io/react_list-of-goods-hooks/


Evgeniy45 (changes requested):
I can`t see if it works, because your demo link does not work
Please fix this problem and i will approve your task ;)

